PR #34, opened 2020-05-19
Fix bug in landscape notebook

kevinsung:
The fix was in line 597:
changed
```python
min_c, max_c = lowest_and_highest_energy(get_problem_graph('HardwareGridProblem'))
```
to
```python
min_c, max_c = lowest_and_highest_energy(get_problem_graph(problem_type))
```
then reran the notebook.


mpharrigan:
I wonder how hard it would be to make the run completely deterministic to reduce some of the churn in filled-in notebooks

PR #36, opened 2020-05-21
Add MGD and optimize submodule

kevinsung:
Split off from #35 . The comments there are already addressed here.


recirq/optimize/minimize_test.py
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import numpy as np
+
+import recirq
+
+
+def sum_of_squares(x):
+    return np.sum(x**2)
```
mpharrigan, line 21: `.item()`

Issue #23, opened 2020-04-10
Pytket dependency issues


mpharrigan:
pytket has a lot of unnecessary dependencies which also cause the dependency resolver to either deadlock or overwrite your version of cirq with an old one. For the CI, we install pytket with `pip install pytket==0.1.6 --no-deps` 

1. Document this quirk
2. Upgrade to newer pytket and hopefully fix the underlying issue? xref #12

mpharrigan:
this was fixed by #38

Issue #12, opened 2020-04-07
Old version of pytket does not install on py3.8


mpharrigan:
1. Use newer pytket
2. Re-enable python 3.8 build

mpharrigan:
The old version of pytket also fails on the latest version of Cirq

mpharrigan:
I couldn't easily figure out how to modify it to use the latest version of pytket. In particular the routing method doesn't return the initial and final mapping.

mpharrigan:
I tried following https://github.com/CQCL/pytket/issues/15 but am now running into issues with their provided "cirq to device" functionality using Cirq `SerializableDevice`.

When using bristlecone `XMonDevice` I was also running into a confusing error message from pytket about an improper qubit/node id

mpharrigan:
I think I've found success porting it to their "compilation unit" framework and my own cirq-to-tket device conversion. PR forthcoming

mpharrigan:
Fixed by #38

PR #38, opened 2020-06-18
Refresh pytket routing code

mpharrigan:
- Use the new compilation unit framework which allows using the latest version of pytket #23 
 - new doc notebook that walks through the bookkeeping
 - testing enabled on py38 #12

### DIFF
--- a/recirq/qaoa/placement.py
+++ b/recirq/qaoa/placement.py
@@ -1,23 +1,19 @@
 import os
 from collections import defaultdict
 from functools import lru_cache
-from typing import Callable, Dict, List, Optional, Tuple, Iterable
+from typing import Callable, Dict, List, Optional, Tuple
 
 import networkx as nx
 import numpy as np
+import pytket
+import pytket.cirq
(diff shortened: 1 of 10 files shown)


recirq/qaoa/placement.py
```diff
@@ -41,6 +37,35 @@ def calibration_data_to_graph(calib_dict: Dict) -> nx.Graph:
...
+    arc = pytket.routing.Architecture(
+        list(_qubit_index_edges(device, qubit_to_index))
+    )
+    return pytket.device.Device({}, {}, arc)
+
+
+def _tk_to_i(tk):
+    """Extract an integer from pytket nodes."""
+    i = tk.index
+    assert len(i) == 1, i
```
kevinsung, line 65: Just `assert len(i) == 1`?

mpharrigan: In the off chance that it isn't 1, it's nice to have the actual value immediately printed in the exception to jump-start debugging

PR #41, opened 2020-07-13
Fix setup.py to read from requirements.txt

yashk2810:


yashk2810:
@lamberta @mpharrigan PTAL

lamberta:
Need this to install requirements with the package (using git install since package not on pypi):

```
$ pip install git+https://github.com/quantumlib/ReCirq
```

lamberta:
cc: @dstrain115

mpharrigan:
This is ok for now, but as we collect more example applications; each with their own set of bespoke dependencies, I'd like to have more fine-tuned dependency management so `pip install recirq` doesn't download everything under the sun

yashk2810:
> doesn't download everything under the sun

Sure you can always use `extra_requires` for that: https://github.com/tensorflow/docs/blob/master/setup.py#L60

PR #48, opened 2020-07-24
add the model based policy gradient method

JiahaoYao:


JiahaoYao:
how to overcome this?

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
+# Copyright 2018 The Cirq Developers
```
mpharrigan, line 1: nit: 2020

recirq/optimize/mpg_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert isinstance(result.x, np.ndarray)
+    assert isinstance(result.fun, float)
+    assert result.nit == 15
+
+
+def test_model_policy_gradient_limited_evaluations():
+    x0 = np.random.randn(10)
+    sample_radius = 1e-1
+    rate = 1e-1
```
mpharrigan, line 85: sample_radius and rate are unused. were they supposed to be used for something?

recirq/optimize/mpg_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+def sum_of_squares(x):
+    return np.sum(x**2).item()
+
+
+def test_model_policy_gradient():
+    x0 = np.random.randn(5)
+    result = model_policy_gradient(sum_of_squares,
+                                    x0,
+                                    learning_rate= 1e-1,
```
mpharrigan, line 28: formatting: omit the space after `=`

recirq/optimize/mpg_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def sum_of_squares(x):
+    return np.sum(x**2).item()
+
+
+def test_model_policy_gradient():
+    x0 = np.random.randn(5)
+    result = model_policy_gradient(sum_of_squares,
+                                    x0,
```
mpharrigan, line 27: formatting: align these indents

recirq/optimize/mpg_test.py
```diff
@@ -0,0 +1,101 @@
...
+import numpy as np
+
+from recirq.optimize.mpg import model_policy_gradient
+
+
+def sum_of_squares(x):
+    return np.sum(x**2).item()
+
+
+def test_model_policy_gradient():
```
mpharrigan, line 24: these tests take ~45 seconds on my machine, which is too long for a unit test. Can we provide a more concise check of the functionality?

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+        [("poly", PolynomialFeatures(degree=2)), ("linear_model", linear_model),]
+    )
+    shifted_xs = [(x - xopt) for x in xs]
+    model = model.fit(shifted_xs, ys)
+    return model
+
+
+def clip_by_norm(t, clip_norm):
+    """Clips tensor values to a maximum L2-norm.
+    returns a tensor of the same type and shape as `t` with its values set to:
```
mpharrigan, line 56: nit: newline between the summary line of the docstring and additional description

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+        Returns: 
+            a class of the schedule
+        """
+        self._v = value
+
+    def value(self, t):
+        """Return the value of the schedule at time step t 
+
+        Args: 
+            t: the time step for the schedule
```
mpharrigan, line 88: note that `t` is ignored here

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+        """Return the value of the schedule at time step t 
+
+        Args: 
+            t: the time step for the schedule 
+
+        Returns: 
+            the schedule value
+        """
+        m = t / self.decay_steps
+        if self.staircase == True:
```
mpharrigan, line 131: remove `== True`, this is a boolean

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+        Args: 
+            t: the time step for the schedule 
+
+        Returns: 
+            the schedule value
+        """
+        return self._v
+
+
+class ExponentialSchedule(object):
```
mpharrigan, line 96: nit: don't need `(object)`

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+        clip_norm: The scalar to give a maximal upper bound for the L2-norm of the tensor.
+
+    Returns:
+        the tensor with the same shape of the input, but with norm clipped.
+    """
+    l2_norm = np.linalg.norm(t, 2)
+    return t * clip_norm / np.maximum(l2_norm, clip_norm)
+
+
+class ConstantSchedule(object):
```
mpharrigan, line 70: nit: don't need `object`.

Consider making this a `dataclass` with `frozen=True`, especially if you're going to use it as a default argument for a function (default arguments should not be mutable)

recirq/optimize/mpg.py
mpharrigan: this can probably be a dataclass as well

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Args:
+            grad: the gradient computed at the current update.  
+            x: the current value of the parameters. 
+
+        Returns:
+            the updated parameter values. 
+        """
+        if self.step == 0:
+            self.m = np.zeros(len(x))
```
mpharrigan, line 173: you shouldn't initialize class members outside of `__init__`

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+                # use the model for prediction
+                new_ys = model.predict(new_xs - current_x)
+                reward = [-y for y in new_ys]
+            except:
+                pass
+
+        reward = np.array(reward)
+
+        # stats
+        reward_mean = np.mean(reward)
```
mpharrigan, line 349: unused

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+                model = _get_quadratic_model(model_xs, model_ys, x)
+
+                # get samples (from model)
+                z = np.random.randn(batch_size_model, n)
+                new_xs = sigma * z + current_x
+
+                # use the model for prediction
+                new_ys = model.predict(new_xs - current_x)
+                reward = [-y for y in new_ys]
+            except:
```
mpharrigan, line 343: this is too broad of an exception catch. Please only catch errors you anticipate so this doesn't silently swallow unrelated, "real" errors

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+            # safer way without the `SVD` not converging
+            try:
+                for x, y in zip(known_xs, known_ys):
+                    if np.linalg.norm(x - current_x) < radius_coeff * max_radius:
+                        model_xs.append(x)
+                        model_ys.append(y)
+                model = _get_quadratic_model(model_xs, model_ys, x)
+
+                # get samples (from model)
+                z = np.random.randn(batch_size_model, n)
```
mpharrigan, line 337: It's usually useful to have deterministic versions of stochastic functions. Consider accepting a numpy random state as an optional argument. see `cirq.parse_random_state` for a helpful utility function and https://cirq.readthedocs.io/en/stable/generated/cirq.RANDOM_STATE_OR_SEED_LIKE.html#cirq.RANDOM_STATE_OR_SEED_LIKE for more details

recirq/optimize/mpg.py
```diff
@@ -0,0 +1,390 @@
...
+            the schedule value
+        """
+        m = t / self.decay_steps
+        if self.staircase == True:
+            m = np.floor(m)
+
+        return self.learning_rate * self.decay_rate ** m
+
+
+class Adam:
```
mpharrigan, line 137: why is this a class? just make it a function

JiahaoYao:
Hi @mpharrigan, received and thanks for your careful comments! Will fix them as soon as possible!

recirq/optimize/mpg.py
JiahaoYao: Good point! Thanks!

recirq/optimize/mpg.py
JiahaoYao: agreed!

recirq/optimize/mpg.py
JiahaoYao: make it narrow

recirq/optimize/mpg_test.py
JiahaoYao: lower the criterion

recirq/optimize/mpg.py
mpharrigan: I meant `Adam`. `ExponentialSchedule` makes sense as a class since it closes over the hyperparameters (learning rate et. al.) and follows the (implicit) Schedule interface

recirq/optimize/mpg.py
```diff
@@ -67,56 +74,31 @@ def clip_by_norm(t, clip_norm):
     return t * clip_norm / np.maximum(l2_norm, clip_norm)
 
 
-class ConstantSchedule(object):
```
mpharrigan: Why did you remove `ConstantSchedule`?

recirq/optimize/mpg.py
```diff
@@ -128,64 +110,60 @@ def value(self, t):
...
         return self.learning_rate * self.decay_rate ** m
 
 
-class Adam:
-    def __init__(
-        self, lr_schedule=ConstantSchedule(0.001), b1=0.9, b2=0.999, eps=10 ** -8
-    ):
-        """The optimizer Adam: a method for stochastic gradient descent  
+def Adam_update(
+    grad,
```
mpharrigan, line 120: type annotations would help here

recirq/optimize/mpg.py
```diff
@@ -128,64 +110,60 @@ def value(self, t):
...
 
         return self.learning_rate * self.decay_rate ** m
 
 
-class Adam:
-    def __init__(
-        self, lr_schedule=ConstantSchedule(0.001), b1=0.9, b2=0.999, eps=10 ** -8
-    ):
-        """The optimizer Adam: a method for stochastic gradient descent  
+def Adam_update(
```
mpharrigan, line 119: I know the original authors style it as "Adam", but I think we should follow PEP8 here and call the function `adam_update`. 

An additional question: do we expect users to use this function or will it always be called from `model_policy_gradient`? That is -- can we make this private (by naming it `_adam_update`)?

recirq/optimize/mpg.py
JiahaoYao: oh, because it is not used in the code? Shall I put it back?

recirq/optimize/mpg.py
JiahaoYao: good point!

recirq/optimize/mpg.py
JiahaoYao: yes, that is right. I have fixed that.

recirq/optimize/mpg.py
JiahaoYao: agreed!

recirq/optimize/mpg.py
JiahaoYao: so, the constant schedule is supposed to ignore the argument `t` because it maps everything to the constant. Can I use `t` as the argument?

recirq/optimize/mpg.py
mpharrigan: Well it was there before so I was wondering why you removed it :) It's up to you and what use cases you want to support

JiahaoYao: ha, I find the schedule is not exposed to the "outside". so, previous constant schedule can not be used. 
But, in practice, it always good to use the decay schedule. If anyone finds the constant schedule is better, I can then put it back.

JiahaoYao:
> Looks great! thanks again. If you do want to put constantschedule back in, feel free to do so in a follow-up pr

Yes, thanks for approving that!

PR #50, opened 2020-08-03
Initial addition of quantum chess libraries

dstrain115:
- Includes code for translating quantum chess moves into circuits,
decomposing into hardware gates, and mapping qubits.
- Adds scipy (at least temporarily) into requirements.  It is only
used for a few tests though, so we could change it.

Does not yet include documentation or experiment scripts.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,6 @@ scikit-learn
 networkx
 pytket
 pytket-cirq
+
+# quantum chess, only needed for tests
+scipy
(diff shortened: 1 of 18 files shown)


requirements.txt
```diff
@@ -10,3 +10,6 @@ scikit-learn
 networkx
 pytket
 pytket-cirq
+
+# quantum chess, only needed for tests
+scipy
```
mpharrigan, line 15: scipy is a requirement for Cirq, so this is fine

PR #52, opened 2020-08-11
Merge 'docs' branch into 'master'

lamberta:
Mostly file moves and renames. Some Markdown fixes.

Some Sphinx-specific files moved into `dev_tools/docs/sphinx/` —just in case you need them and can probably delete from repo when you are ready.

PR #55, opened 2020-08-12
Disable sphinx doc build

mpharrigan:
xref gh-52


mpharrigan:
No, sphinx was still working fine. This is for the switchover to the new docs system: #52

PR #59, opened 2020-08-12
Adding scripts to quantum chess

dstrain115:
- Adds 'experiments' to quantum_chess.
- First is an interactive mode where you can type quantum moves and
see how the board develops.
- Second is a batch mode that can execute a sequence of moves on
a quantum board and display the final result.

- Also adds 54-qubit sycamore simulators to engine_utils.

Documentation is still pending.


recirq/quantum_chess/experiments/batch_moves.py
```diff
@@ -0,0 +1,74 @@
...
+        b.load_fen(fen_position)
+    else:
+        b.reset()
+    print(f'Applying moves to board...')
+    apply_moves(b, moves)
+    print(b)
+
+
+if __name__ == '__main__':
+    filename = sys.argv[1] if len(sys.argv) > 1 else None
```
mpharrigan, line 69: you don't want to use argparse?

I'd also suggest wrapping up this code in a simple `parse()` method:

```python
def parse():
  filename = ..
  processor_name = ...
  return main_loop(...)

if __name__ == '__main__':
  parse()
```

because otherwise `filename` `position` and `processor_name` are technically global variables which has a low probability of causing problems elsewhere, but if it does: they're a real head-scratcher

recirq/quantum_chess/experiments/interactive_board.py
```diff
@@ -0,0 +1,73 @@
...
+    if fen_position:
+        b.load_fen(fen_position)
+    else:
+        b.reset()
+    while True:
+        print('')
+        print(b)
+        print('')
+        try:
+            in_str = next(sys.stdin).strip()
```
mpharrigan, line 55: why aren't you using `input()`? Instead of `while True` can't you use a `for in_str in sys.stdin`? There's also the `fileinput` https://docs.python.org/3/library/fileinput.html module which would be better

recirq/quantum_chess/experiments/batch_moves.py
dstrain115: Done.  Changed to use argparse.

recirq/quantum_chess/experiments/interactive_board.py
dstrain115: Done.

PR #64, opened 2020-08-26
Allow sphinx building of docs

mpharrigan:
The primary mechanism will be the `docs/` folder and
associated tensorflow-like build system. However, for
the interim we'll support building with sphinx since there's
little overhead.

In particular, sphinx will copy files from `docs/` to
`dev_tools/docs/sphinx`.

This PR also makes the following changes:

 - change copyright notice to bold from `<h6>`. `nbsphinx`
   was getting confused.
 - only install `recirq` in the notebook if it's not already
   installed. It's not nice to overwrite a user's environment!

### DIFF
--- a/dev_tools/docs/sphinx/conf.py
+++ b/dev_tools/docs/sphinx/conf.py
@@ -147,5 +147,18 @@ def env_before_read_docs(app, env, docnames):
         _rmtree_if_exists(f'{qaoa_p1_tasks.DEFAULT_BASE_DIR}/2020-03-tutorial')
 
 
+import subprocess
+
+REPO_DIR = subprocess.Popen(['git', 'rev-parse', '--show-toplevel'],
+                            stdout=subprocess.PIPE).communicate()[0].rstrip().decode('utf-8')
+
+
 def setup(app):
+    FROM = f'{REPO_DIR}/docs'
+    TO = f'{REPO_DIR}/dev_tools/docs/sphinx'
+    from distutils.dir_util import copy_tree  # allows overwriting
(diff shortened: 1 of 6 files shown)


docs/tutorials/data_analysis.ipynb
```diff
@@ -7,7 +7,7 @@
         "id": "aA39B_EcZQCc"
       },
       "source": [
-        "##### Copyright 2020 Google"
+        "**Copyright 2020 Google**"
```
lamberta, line 10: Ah. Yeah we use this format for our search and strip when importing into the site

mpharrigan: Ok, I changed sphinx to do this substitution during build time with a simple regex

dev_tools/docs/sphinx/conf.py
```diff
@@ -147,5 +152,24 @@ def _order(docname):
...
 
 
+import subprocess
+
+REPO_DIR = subprocess.Popen(['git', 'rev-parse', '--show-toplevel'],
+                            stdout=subprocess.PIPE).communicate()[0].rstrip().decode('utf-8')
+
+
+def source_read(app, docname, source):
+    import re
```
balopat, line 162: nit: why not import on top of the file?

mpharrigan: as I was typing it I knew someone would comment on it. The real reason is because I was doing it quickly and didn't want to scroll all the way up to the top! The plausible deniability reason is because this is a highly-customized change, I kept all the non-standard sphinx conf.py stuff into this function

mpharrigan: done

PR #65, opened 2020-08-28
More sphinx docs

mpharrigan:
Following #64 

Also adds back in outputs for long-running notebooks.


docs/qaoa/routing_with_tket.ipynb
```diff
@@ -624,13 +627,12 @@
   "metadata": {
     "colab": {
       "name": "routing_with_tket.ipynb",
-      "private_outputs": true,
+      "private_outputs": false,
```
lamberta, line 630: The Colab naming always confuses me.
Did you mean to set this notebook to `private_outputs: false`? (cc: @yashk2810)

mpharrigan: I think `nbfmt` did this, so any clarity would be helpful!

yashk2810: Yeah, this is confusing.

In the METADATA checks we require private_outputs to be set to True.

mpharrigan: It looks like `nbfmt` (line 246) sets

```python
colab["private_outputs"] = not FLAGS.preserve_outputs
```

What is private outputs?

lamberta: Yeah, chatted offline and will update `nbfmt`. This setting is kind of legacy for the old docs infra. Colab defaults to not including this field so we'll follow. (Previously we used this to determine executability of the notebook, but I've always thought the name was confusing :P)

In Colab, `private_outputs` is a setting for whatever a user can save output cells in this notebook (or will need to create a copy in Colab to enable)

I'll update `nbfmt` to remove this field like the Colab default.

lamberta: Updated `nbfmt` in cl/329538550, can you please update and run again?
We're no longer using the `private_outputs` field for anything so stripping it (like the Colab default). We look for the `"output_type"` to indicate it outputs are saved with the notebook (and, therefore, won't execute).

The `--preserve_outputs` flag has been removed and we now have a `--remove_outputs` flag. By default it will respect the current notebook setting (i.e. not remove outputs if they exist).

mpharrigan: done

PR #84, opened 2020-10-21
Re-organize docs

mpharrigan:
- New wording for index
 - Move things to "concepts" heading and subdirectory
 - Move Hartree Fock notebook into docs/
 - Add FHM notebooks to _toc

Outstanding:
 - More wording on the index landing page
 - New HFVQE tutorials
 - Changes to the FHM tutorial so they execute successfully in colab


lamberta (approved):
I think this makes sense and for a better url structure. But you're basically using `concepts/` as a `guide/` section and this might be a bit confusing as you collect more docs and contributors submit docs---this could turn into a grab bag and take some discipline to organize.
I guess one way to think about it: will this be the "ReCirq user guide"? If so, I'd go with `guide/` to be consistent with the rest of the site

mpharrigan:
changed to "guide"

lamberta (approved):

PR #86, opened 2020-10-22
Docs nav update

lamberta:
* Update nav to use `/cirq/experiments/`.
* File rename to `hfvqe/odd_qubit.ipynb`.
* Update button links.
* Run `nbfmt` to clear some extra Colab fields.

Some new notebooks need the buttons/license template, but will add that later.


mpharrigan (approved):
Thank you!

PR #89, opened 2020-10-28
Index pages for doc sections

mpharrigan:
We need overview/index/landing pages for each of the sub-sections to serve as a stable base for other content to link to. We need these pages to exist ASAP; so please raise an issue if anything is catastrophically wrong with the stubbed out content I put in. Otherwise, these pages can be refined (or even replaced with ipython notebooks `index.ipynb`!) later.

@ncrubin - please verify changes to HFVQE are acceptable

@cantwellc @dstrain115 - please verify quantum chess overview is acceptable

@mrwojtek - please verify fermi-hubbard overview is acceptable

@lamberta @karlunho - check TOC, general structure

PR #90, opened 2020-10-28
Notebook format, buttons, and setup

lamberta:
Add license, buttons, and setup to install packages.
Apply `nbfmt` to notebooks.


mpharrigan:
Can we do nbfmt with 1 space to prevent these huge diffs?

lamberta:
> Can we do nbfmt with 1 space indent to prevent these huge diffs?

Right now the ReCirq notebooks are a mix of 1 and 2 spaces. This PR makes them all 2 to make it consistent, but I can make them all 1, if you prefer. You'll still get some large diffs in this PR but shouldn't going forward. I believe Cirq went with 1, as well.

lamberta:
Will need to rebase with https://github.com/quantumlib/ReCirq/pull/89

mpharrigan:
Let's do all 1 space indents for sure.

If possible, can you have the automated nbfmt changes in a separate commit so I can tell if it's just the indent level changing vs. any additional changes (buttons, setup, ...)

lamberta:
Yeah, sounds good. Lemme get the nav in order and then I'll come back to this PR. I'll split up the nbfmt and the button changes to make it easier to review

Issue #83, opened 2020-10-21
Colab notebooks need `pip install`


mpharrigan:
When running documentation notebooks in Colab, the runtime does not come with (Re)Cirq installed. Luckily, you can execute shell commands by prepending a `!` to a cell:

```
!pip install recirq
```

However

1) ReCirq isn't released to PyPI yet (and sometimes notebooks will be documenting unreleased functionality)
2) This might clobber developer / custom / conda installations of the packages if the user *isn't* running in Colab.

Therefore, all the notebooks should have 

```
try:
    import recirq
except ImportError:
    !pip install --quiet git+https://github.com/quantumlib/ReCirq
```

as the install step.

mpharrigan:
xref https://github.com/quantumlib/Cirq/pull/3431

mpharrigan:
This is done

PR #96, opened 2020-11-05
Polish Fermi-Hubbard experiment example.

rmlarose:
- Add Hamiltonian and charge/spin densities.
- Add frontmatter (run on Colab, download notebook, etc.)
- Used standard install/import scheme.
- Remove unused imports.
- Add a bit more explanatory text.
- Minor typo fixes, changes, etc.


mpharrigan:
You seem to have committed all the outputs. Was this intentional?

rmlarose:
> You seem to have committed all the outputs. Was this intentional?

Indeed. I thought this was the preferred method esp. for devsite staging (cc: @balopat). If not, it's no problem to clear the outputs.

mpharrigan:
I think we want to clear outputs unless the notebook is long running or can't be run standalone in colab

mpharrigan (approved):

mpharrigan:
@mrwojtek does this look good to you?

docs/fermi_hubbard/experiment_example.ipynb
```diff
@@ -1,12 +1,65 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "8d9dec56909c"
+   },
+   "source": [
+    "<table class=\"tfo-notebook-buttons\" align=\"left\">\n",
+    "  <td>\n",
+    "    <a target=\"_blank\" href=\"https://www.example.org/cirq/experiments/fermi_hubbard/experiment_example\"><img src=\"https://www.tensorflow.org/images/tf_logo_32px.png\" />View on QuantumLib</a>\n",
```
mrwojtek, line 51: This points to a dead domain.

docs/fermi_hubbard/experiment_example.ipynb
```diff
@@ -15,7 +68,90 @@
     "id": "9ec578c1796c"
    },
    "source": [
-    "This notebook demonstrates how to execute a single instance of Fermi-Hubbard experiment on Google processor."
+    "This notebook demonstrates how to define, execute, and plot the results of a single instance of [the Fermi-Hubbard experiment](https://arxiv.org/abs/2010.07965). We show how to run the experiment using a Cirq simulator and a quantum processor through Google's Quantum Computing Service.\n",
+    "\n",
+    "The Fermi-Hubbard model on a one-dimensional lattice of $L$ sites with open boundary conditions is defined by the Hamiltonian\n",
+    "\n",
+    "$$\n",
+    "H = - J \\sum_{j = 1}^{L - 1} \\sum_{\\nu = \\uparrow, \\downarrow} c_{j, \\nu}^\\dagger c_{j + 1, \\nu} + \\text{h.c.} + U \\sum_{j = 1}^{L} n_{j, \\uparrow} n_{j, \\downarrow} + \\sum_{j = 1}^{L} \\sum_{\\nu = \\uparrow, \\downarrow} \\epsilon_{j, \\nu} n_{j, \\nu} \n",
```
mrwojtek, line 76: What we are really supporting in this experiment is an extended Hamiltonian with a next-neighbor interaction terms V = n_{j, \\nu} n_{j + 1, \\nu}. This is mentioned on page 2 of the Fermi-Hubbard paper. We didn't include that in the equation (1) of the paper because of the clarity.

I think, this codebase is not only the Fermi-Hubbard paper and it might be good to include this full Hamiltonian description. This is not the usual Fermi-Hubbard model though. This V term is used in "Parasitic-controlled-phase" section below for example.

mrwojtek: The code also supports site-dependent U, J and V coefficients (something like U_j, J_j, and V_j). In the paper we simplified it because we only used site-dependence for V term but this codebase supports more general Hamiltonians.

mrwojtek (approved):
I really appreciate those improvements, the document is in much better shape now!

docs/fermi_hubbard/experiment_example.ipynb
rmlarose: This is true for all tutorials currently, e.g. [QAOA](https://github.com/quantumlib/ReCirq/blob/master/docs/qaoa/example_problems.ipynb) and [Cirq basics](https://github.com/quantumlib/Cirq/blob/master/docs/tutorials/basics.ipynb). I think the plan is to update the links once the devsite is public - do you know more @mpharrigan?

rmlarose: Actually, the "View on QuantumLib" links do not show up on the devsite.

mpharrigan: cc @lamberta , but I think @rmlarose is correct: we'll update once we have something to point to!

lamberta: Correct. We'll need to do a link sweep post-devsite (also update the icons). But this first button won't display on the site (but will in GitHub and Colab)

lamberta:
> I think we want to clear outputs unless the notebook is long running or can't be run standalone in colab

Yeah. Preferred to *not* save outputs so we can test the notebook. But if for some reason we can't test it (I dunno, maybe some kind of Google Cloud interaction that can't run in our VM), then you can save the notebooks with outputs and we'll publish as-is ... but you'll lose the ability to test them.

docs/fermi_hubbard/experiment_example.ipynb
```diff
@@ -667,9 +931,22 @@
   },
   "kernelspec": {
    "display_name": "Python 3",
+   "language": "python",
    "name": "python3"
+  },
+  "language_info": {
```
lamberta, line 937: Can you run this through [nbfmt](https://github.com/tensorflow/docs/blob/master/tools/tensorflow_docs/tools/) please? This normalizes the JSON and removes extra metadata which makes it easier to review diffs later.

```
$ python3 -m pip install -U --user git+https://github.com/tensorflow/docs

$ python3 -m tensorflow_docs.tools.nbfmt --indent=1 notebook.ipynb
```

Quantum folks prefer `--indent=1`

docs/fermi_hubbard/experiment_example.ipynb
rmlarose: Thanks a lot @mrwojtek for these comments. I added a note about the parasitic controlled phase as well as supporting site-dependent coefficients. I kept the equation of the Hamiltonian the same as per the paper, but it's also possible to include the cphase-induced interaction term explicitly. Let me know what you think.

mrwojtek: I think it's perfectly fine to put all this information in the text and not in the main equation.

I tend to think about the V term more like an ability to simulate extended physics, less as a necessary evil caused by our hardware, i.e., this is not only used for simulation on classical computer. If you introduce non-zero V terms and run it on a hardware, the circuits will have nearest-neighbor CPHASE gates and this extended Hubbard model will be actually simulated on a quantum device. Could you reformulate this more in a way that this is an extension to the model that we support, with less impact on the details caused by the device behavior?

mpharrigan:
We good? @mrwojtek

mrwojtek:
Well, the description is slightly confusing right now because it suggests that V term is only used to compensate for hardware errors. The V terms can also be used to simulate the extended Hubbard models, even on hardware.

rmlarose:
Rephrased description of extended Hamiltonian to sound less negative :) PTAL @mrwojtek

mrwojtek:
Thank you Ryan, this is much more diplomatic right now, I have no further comments ;)

mpharrigan:
cc #83

PR #98, opened 2020-11-18
Add code for out-of-time-order correlator experiments

XiaoMiQC:
Add functions and examples for performing an out-of-time-order correlator (OTOC) experiment


recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+        cycle: int,
+        qubit_order: List[Any],
+        sq_gates: np.ndarray,
+        interaction_sequence: Sequence[Set[Tuple[Any, Any]]],
+        num_replaced_gates: int,
+        rand_seed: int = None,
+        to_cifford_only: bool = False,
+        rand_seed_replaced_gates: int = None
+) -> np.ndarray:
+    r"""Replace a select number of single-qubit gatrs with other random gates.
```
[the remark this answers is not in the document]
XiaoMiQC: Done

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+            echo_indices=echo_indices_back)
+
+    butterfly_ops = [None, cirq.X, cirq.Y, cirq.Z]
+    prep_phases = [0.0, 1.0]
+    meas_phases = [0.0, 1.0]
+    otoc_circuits = []
+
+    for b_ops in butterfly_ops:
+        for p_prep in prep_phases:
+            for p_meas in meas_phases:
```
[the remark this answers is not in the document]
XiaoMiQC: Done

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+
+                circuit_full.append(cirq.PhasedXPowGate(
+                    phase_exponent=p_meas, exponent=0.5)(ancilla))
+                circuit_full.append(cirq.measure(ancilla, key='z'),
+                                    strategy=cirq.InsertStrategy.NEW)
+                otoc_circuits.append(circuit_full)
+    return otoc_circuits
+
+
+def replace_gates(
```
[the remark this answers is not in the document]
XiaoMiQC: This is a function that replaces a certain number of gates in a Clifford circuit with non-Clifford gates. It's not quite used here since iSWAP is not supported (sqrt(iSWAP) is not a Clifford so this function doesn't serve any purpose for circuits constructed with currently available gates. I will modify the doc-string to mention this fact.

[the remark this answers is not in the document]
XiaoMiQC: Done. Replaced with add_non_cliffords

recirq/otoc/otoc_circuits.py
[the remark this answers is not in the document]
XiaoMiQC: num_replaced_gates controls how many SQ gates get replaced with nonCliffords. The two random seed args control the locations and rotation axes of the replacing gates. The other args are all meant to specify a lightcone structure for the OTOC circuit to ensure the nonClifford rotations get placed into places where they produce non-trivial change in the OTOC

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+        meas_qubit: Any,
+        butterfly_qubits: Union[Any, Sequence[Any]],
+        cycle: int,
+        qubit_order: List[Any],
+        sq_gates: np.ndarray,
+        interaction_sequence: Sequence[Set[Tuple[Any, Any]]],
```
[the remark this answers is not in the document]
XiaoMiQC: I tried to make it more general here such that "qubits" can be represented by any object. For example, they can be either specified as GridQubits or just tuples like (3, 4).

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+    Returns:
+        A list of 16 cirq.Circuits. The first 4 circuits are OTOC circuits
+        with the butterfly operators being I (i.e. no perturbation between U
+        and $U^\dagger$). Circuits 4 to 8, 8 to 12 and 12 to 16 correspond to
+        the butterfly operators being X, Y and Z, respectively. For each
+        case, the OTOC measurement result is $(p_0 - p_1 - p_2 + p_3) / 2$,
+        where $p_i$ is the excited state probability of the $i^\text{th}$
+        circuit for the corresponding butterfly operator.
+    """
+    if cliffords and z_only:
```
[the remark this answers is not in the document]
XiaoMiQC: Done. Added single-line comments for different parts of the function.

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+    return new_sq_gates
+
+
+def _compile_moments(
+        ops_steps: Sequence[Set[Tuple[cirq.GridQubit, cirq.GridQubit]]],
+        ops_dict: Union[_CORRECTION_FILE, List[_CORRECTION_FILE]],
+        layer_by_layer_cal: bool = False
+) -> List[Union[cirq.Moment, Sequence[cirq.Moment]]]:
+    moment_list = []
+    num_slices = 3
```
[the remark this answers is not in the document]
XiaoMiQC: You are right, this is actually not necessary. I deleted it.

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+                moment_list.append([cirq.Moment(m) for m in ops_list])
+    return moment_list
+
+
+def _compile_interactions(first_qubit: Any,
+                          butterfly_qubit: Union[Any, Sequence[Any]],
+                          cycle: int,
+                          interaction_sequence: Sequence[Set[Tuple[Any, Any]]]
+                          ) -> Tuple[List[Set[Tuple[Any, Any]]],
+                                     List[Set[Tuple[Any, Any]]]]:
```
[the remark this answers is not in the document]
XiaoMiQC: Added doc-string

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+                        from_right: bool,
+                        interaction_layers: Sequence[Set[Tuple[Any, Any]]],
+                        multiple_butterflies: bool = False
+                        ) -> List[Set[Any]]:
+    num_configs = len(interaction_layers)
+    int_seq = [interaction_layers[c % num_configs] for c in range(cycle)]
+    if reverse ^ from_right:
+        int_seq.reverse()
+
+    if multiple_butterflies is True:
```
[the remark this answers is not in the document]
XiaoMiQC: Yeah this is a little clunky, but isintance(butterfly_qubits, Sequence) doesn't quite work as sometimes the qubits are represented by Tuples like (3, 5) that are instances of Sequence themselves.

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+            qubits, [cycle], moments_for, sq_rand_nums=rand_nums,
+            z_only=z_only, ancilla=ancilla, cycles_per_echo=cycles_per_echo)
+        circuits_backward, _ = build_xeb_circuits(
+            qubits, [cycle], moments_back, sq_rand_nums=rand_nums_rev,
+            z_only=False, ancilla=ancilla, cycles_per_echo=cycles_per_echo,
+            reverse=True)
+
+    else:
+        light_cone_u = _extract_light_cone(
+            butterfly_qubits, cycle, False, True, interaction_sequence,
```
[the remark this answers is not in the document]
XiaoMiQC: Done. Rearranged the keyword arguments. reverse and from_right are always explicitly stated whenever the function is invoked

recirq/otoc/utils.py
```diff
@@ -0,0 +1,253 @@
...
+                   delta_minus_diag: float, delta_minus_off_diag: float
+                   ) -> np.ndarray:
+    """Converts five phases to a 2x2 FSIM unitary.
+
+    Args:
+        theta: The swap angle.
+        phi: The conditional phase.
+        delta_plus: A single-qubit phase.
+        delta_minus_diag: A single-qubit phase.
+        delta_minus_off_diag: A single-qubit phase.
```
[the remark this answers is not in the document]
XiaoMiQC: Thanks. I will add this in the future.

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,499 @@
...
+        num_replaced_gates: The total number of single-qubit gates to replace.
+        rand_seed: The random seed for the locations of the replaced gates.
+            If unspecified, no random seed will be used.
+        to_cifford_only: If True, the replaced gates will be drawn randomly
+            from +/-X/2 and +/-Y/2.
+        rand_seed_replaced_gates: The random seed for the gate replacements.
+
+    Returns:
+        An np.ndarray with dimensions len(qubit_order) $\times$ cycle.
+    """
```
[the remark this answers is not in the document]
XiaoMiQC: Done.

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,723 @@
...
+    cirq.Y ** 0.5, cirq.PhasedXPowGate(phase_exponent=0.75, exponent=0.5),
+    cirq.X ** -0.5, cirq.PhasedXPowGate(phase_exponent=-0.75, exponent=0.5),
+    cirq.Y ** -0.5, cirq.PhasedXPowGate(phase_exponent=-0.25, exponent=0.5),
+]
+_rot_mats = [cirq.unitary(r) for r in _rot_ops]
+_fsim_angle_labels = ['theta', 'delta_plus', 'delta_minus_off_diag',
+                      'delta_minus_diag', 'phi']
+
+
+def default_interaction_sequence(
```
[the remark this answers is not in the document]
XiaoMiQC: Not much in the current code. I think it would be good to retain this as it is convenient for breaking gates in 2D into different layers for parallel XEB.

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,723 @@
...
+        if len(gate_set) != 0:
+            l_final.append(gate_set)
+
+    return l_final
+
+
+def build_xeb_circuits(qubits: Sequence[cirq.GridQubit],
+                       cycles: Sequence[int],
+                       benchmark_ops: Sequence[Union[
+                           cirq.Moment, Sequence[cirq.Moment]]] = None,
```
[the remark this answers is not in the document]
XiaoMiQC: Done.

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+
+from recirq.otoc.otoc_circuits import build_otoc_circuits
+
+
+def main():
+    # Specify project ID and processor name.
+    project_id = ''
+    engine = cirq.google.Engine(project_id=project_id,
+                                proto_version=cirq.google.ProtoVersion.V2)
+    processor_name = 'rainbow'
```
mpharrigan, line 29: Use `cirq.google.get_engine_sampler` which will use the GOOGLE_CLOUD_PROJECT environment variable, if not otherwise provided. This will also return a `QuantumEngineSampler`, which is how we want users to be executing programs since it can be substituted for a simulator

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+    # Specify qubits to measure. The qubits must form a line. The ancilla qubit
+    # must be connected to the first qubit.
+    qubit_locs = [(3, 3), (2, 3), (2, 4), (2, 5), (1, 5), (0, 5), (0, 6),
+                  (1, 6)]
+    qubits = [cirq.GridQubit(*idx) for idx in qubit_locs]
+    ancilla_loc = (3, 2)
+    ancilla = cirq.GridQubit(*ancilla_loc)
+    num_qubits = len(qubits)
+
+    # Specify how the qubit interact. For the 1D chain in this example, only two
```
mpharrigan, line 40:
```suggestion
    # Specify how the qubits interact. For the 1D chain in this example, only two
```

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+    # layers of two-qubit gates (CZ is used here) is needed to enact all
+    # interactions.
+    int_sets = [
+        {(qubits[i], qubits[i + 1]) for i in range(0, num_qubits - 1, 2)},
+        {(qubits[i], qubits[i + 1]) for i in range(1, num_qubits - 1, 2)}]
+
+    forward_ops = {(qubit_locs[i], qubit_locs[i + 1]): cirq.Circuit([cirq.CZ(
```
mpharrigan, line 47: the wrapping is weird here. Consider using the `black` formatter like we use in Cirq with a line-length of 100

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+
+    reverse_ops = {(qubit_locs[i], qubit_locs[i + 1]): cirq.Circuit([cirq.CZ(
+        qubits[i], qubits[i + 1])]) for i in range(num_qubits - 1)}
+
+    # Build two random circuit instances (each having 12 cycles).
+    circuit_list = []
+    cycles = range(12)
+    num_trials = 2
+    for i in range(num_trials):
+        np.random.seed(i)
```
mpharrigan, line 58: why are you setting the seed inside the loop? Consider doing it once outside the loop. For this example, it might be most appropriate to set the seed globally at the top of the file.

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+        circuits_i = []
+        for cycle in cycles:
+            circuits_ic = []
+            for k, q_b in enumerate(qubits[1:]):
+                circs = build_otoc_circuits(
+                    qubits, ancilla, cycle, int_sets, forward_ops=forward_ops,
+                    reverse_ops=reverse_ops, butterfly_qubits=q_b,
+                    cycles_per_echo=2, sq_gates=rand_nums, use_physical_cz=True)
+                if k == 0:
+                    circuits_ic.extend(circs[0:8])
```
mpharrigan, line 69: can you explain this conditional?

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+        circuit_list.append(circuits_i)
+
+    # Measure the OTOCs of the two random circuits.
+    results = []
+    for i, circuits_i in enumerate(circuit_list):
+        results_i = np.zeros((num_qubits, len(cycles)))
+        for c, circuits_ic in enumerate(circuits_i):
+            print('Measuring circuit instance {}, cycle {}...'.format(i, c))
+            stats = int(2000 + 10000 * (c / max(cycles)) ** 3)
+            sweep_params = [{} for _ in range(len(circuits_ic))]
```
mpharrigan, line 82: use `sampler.run_batch` and this will be the default (i.e. no need to construct)

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+            job = engine.run_batch(
```
mpharrigan, line 83: use `sampler.run_batch`

recirq/otoc/otoc_example.py
```diff
@@ -0,0 +1,109 @@
...
+                repetitions=stats,
+                processor_ids=[processor_name],
+                gate_set=cirq.google.FSIM_GATESET)
+            for d in range(num_qubits):
+                p = np.mean(job.results()[4 * d].measurements['z'])
+                p -= np.mean(job.results()[4 * d + 1].measurements['z'])
+                p -= np.mean(job.results()[4 * d + 2].measurements['z'])
+                p += np.mean(job.results()[4 * d + 3].measurements['z'])
+                results_i[d, c] = 0.5 * p
+            results.append(results_i)
```
mpharrigan, line 95: is this indented correctly? You initialize `results_i` once per `circuits_i` but append it every `circuits_ic`.

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,571 @@
...
+import itertools
+from typing import Sequence, Union, Tuple, List, Dict, Set, Any, Optional
+
+import cirq
+import numpy as np
+
+from recirq.otoc.parallel_xeb import build_xeb_circuits
+from recirq.otoc.utils import cz_to_sqrt_iswap
+
+_CORRECTION_FILE = Dict[Tuple[Tuple[int, int], Tuple[int, int]], cirq.Circuit]
```
mpharrigan, line 25: What is this? Why is it called a "file"? Please document here or in the argument docs for `build_otoc_circuits`

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,571 @@
...
+            butterfly qubit(s). Has no effect if light_cone_filter is False.
+        idle_echo_seed: The seed for the random spin-echo gates for qubits
+            outside the light-cones of the butterfly qubit(s). If
+            unspecified, no seed will be used. Has no effect if
+            light_cone_filter is False.
+
+    Returns:
+        A list of 16 cirq.Circuits. The first 4 circuits are OTOC circuits
+        with the butterfly operators being I (i.e. no perturbation between U
+        and $U^\dagger$). Circuits 4 to 8, 8 to 12 and 12 to 16 correspond to
```
mpharrigan, line 122: There's a ton of structured coded into the indices in this one flat list. Consider creating a small, self-documenting dataclass to clear this up. I think I made a note when I was confused about this reading through the example

```python
@dataclass
OTOCCircuits:
  butterfly_I: List[cirq.Circuits]
  butterfly_X: List[
  butterfly_Y ...
```

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,571 @@
...
+        raise ValueError('cliffords and z_only cannot both be True')
+
+    num_ops = len(interaction_sequence)
+    num_qubits = len(qubits)
+    multi_b = True if np.ndim(butterfly_qubits) == 1 else False
+
+    # Specify the random single-qubit gates (rand_num) in term of random
+    # np.ndarray. The inverse of the single-qubit gates (rand_nums_rev) is
+    # also specified.
+    if random_seed is not None:
```
mpharrigan, line 138: Use `cirq.value.parse_random_state` which lets the user supply a `RandomState` instance. It returns a `RandomState` instance. Then replace all usages of `np.random` with the returned object (it has all the same methods)

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,571 @@
...
+
+    butterfly_ops = [None, cirq.X, cirq.Y, cirq.Z]
+    prep_phases = [0.0, 1.0]
+    meas_phases = [0.0, 1.0]
+    otoc_circuits = []
+
+    # Combine $U$ and $U^\dagger$ with other gates related to SPAM to
+    # complete the OTOC circuits.
+    for b_ops, p_prep, p_meas in itertools.product(
+            butterfly_ops, prep_phases, meas_phases):
```
mpharrigan, line 236: Weird wrapping. We can use 100 character line-length

recirq/otoc/otoc_circuits.py
```diff
@@ -0,0 +1,571 @@
...
+        for pair in int_list:
+            if (pair[0] in reverse_cone_0[i]) and (
+                    pair[1] in reverse_cone_0[i]):
+                new_set.add(pair)
+        reverse_seq.append(new_set)
+
+    return forward_seq, reverse_seq
+
+
+def _extract_light_cone(butterfly_qubits: Union[Any, Sequence[Any]],
```
mpharrigan, line 524: To these private functions at the end, can you add a brief (i.e. one line) docstring giving a very concise summary of what they do and/or how they're used by the "main" functions

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+
+from recirq.otoc.parallel_xeb import build_xeb_circuits, parallel_xeb_fidelities
+from recirq.otoc.utils import save_data
+
+
+def main():
+    # Specify a working directory, project ID and processor name.
+    dir_str = ''
+    project_id = ''
+    engine = cirq.google.Engine(project_id=project_id,
```
mpharrigan, line 28: use `cg.get_engine_sampler`

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+    xeb_configs = [[cirq.Moment([cirq.ISWAP(qubits[i], qubits[i + 1]) ** 0.5
+                                 for i in range(0, num_qubits - 1, 2)])],
+                   [cirq.Moment([cirq.ISWAP(qubits[i], qubits[i + 1]) ** 0.5
+                                 for i in range(1, num_qubits - 1, 2)])]]
+
+    # Specify the number of random circuits in parallel XEB the cycles at which
+    # fidelities are to be measured.
+    num_trials = 20
+    num_cycle_range = range(3, 75, 10)
+    num_cycles = len(num_cycle_range)
```
mpharrigan, line 55: `num_num_cycles`?

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+         range(1, num_qubits - 1, 2)}]
+
+    xeb_configs = [[cirq.Moment([cirq.ISWAP(qubits[i], qubits[i + 1]) ** 0.5
+                                 for i in range(0, num_qubits - 1, 2)])],
+                   [cirq.Moment([cirq.ISWAP(qubits[i], qubits[i + 1]) ** 0.5
+                                 for i in range(1, num_qubits - 1, 2)])]]
+
+    # Specify the number of random circuits in parallel XEB the cycles at which
+    # fidelities are to be measured.
+    num_trials = 20
```
mpharrigan, line 53: `num_circuits`?

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+    # the data as well as random circuit information.
+    all_bits_sqrtiswap = []
+    all_phases_sqrtiswap = []
+    for xeb_config in xeb_configs:
+        bits_sqrtiswap = []
+        phases_sqrtiswap = []
+        for i in range(num_trials):
+            print(i)
+            circuits, rand_indices = build_xeb_circuits(
+                qubits, num_cycle_range, xeb_config, random_seed=i)
```
mpharrigan, line 67: this isn't how deterministic random numbers is supposed to work. Seed once (either a global call to `np.random.seed()` or construct a `np.random.RandomState(seed)` and pass to this function

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+    num_cycle_range = range(3, 75, 10)
+    num_cycles = len(num_cycle_range)
+
+    # Generate random XEB circuits, measure the resulting bit-strings, and save
+    # the data as well as random circuit information.
+    all_bits_sqrtiswap = []
+    all_phases_sqrtiswap = []
+    for xeb_config in xeb_configs:
+        bits_sqrtiswap = []
+        phases_sqrtiswap = []
```
mpharrigan, line 63: why is this called `phases`? The return variable is named `rand_indices`. The second return value is not documented in `build_xeb_circuits`.

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+        for i in range(num_trials):
+            print(i)
+            circuits, rand_indices = build_xeb_circuits(
+                qubits, num_cycle_range, xeb_config, random_seed=i)
+            phases_sqrtiswap.append(rand_indices)
+            for c in circuits:
+                c.append(cirq.measure(*qubits, key='z'))
+            sweep_params = [{} for _ in range(len(circuits))]
+            job = engine.run_batch(
```
mpharrigan, line 72: use `sampler.run_batch` to simplify execution

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+        all_bits_sqrtiswap.append(bits_sqrtiswap)
+        all_phases_sqrtiswap.append(phases_sqrtiswap)
+
+    # Perform fits on each qubit pair to miminize the cycle errors. The fitting
+    # results are saved to the working directory.
+    fsim_angles_0 = {'theta': -0.25 * np.pi, 'delta_plus': 0,
+                     'delta_minus_off_diag': 0, 'delta_minus_diag': 0,
+                     'phi': 0.0}
+    (fitted_gates_0, corrected_gates_0, fsim_fitted_0, err_0_opt,
+     err_0_unopt) = parallel_xeb_fidelities(
```
mpharrigan, line 89: weird wrapping. Consider using a dataclass for the return value of `parallel_xeb_fidelities`.

```python
@dataclass
XEBResults:
  fitted_gates: ...
  corrected_gates: ...
```

Why do these all have a `_0` in them?

recirq/otoc/parallel_xeb_example.py
```diff
@@ -0,0 +1,100 @@
...
+        qubit_locs, num_cycle_range, all_bits_sqrtiswap,
+        all_phases_sqrtiswap, fsim_angles_0,
+        interaction_sequence=int_layers, gate_to_fit='sqrt-iswap',
+        plot_individual_traces=True, num_restarts=5,
+        plot_histograms=True, save_directory=None)
+
+    save_data(corrected_gates_0, dir_str + '/gate_corrections')
```
mpharrigan, line 96: If the user doesn't modify this example before running, it will try to save the data in the hard drive root `/gate_corrections` absolute path. Please provide a sensible default like the current directory

recirq/otoc/utils.py
```diff
@@ -0,0 +1,253 @@
...
+
+import pickle
+from typing import Any, Optional, Tuple, Dict, List, Sequence
+
+import cirq
+import numpy as np
+from scipy.optimize import curve_fit
+
+
+def save_data(data: Any, file_name: str) -> None:
```
mpharrigan, line 25: pickle is a very bad format for data storage. It is an opaque format that can break very easily between python versions or even if you change the name of a class or function or even if you change a class attribute from a normal attribute to a `@property`. This is probably outside the scope of the current review, but we should open an issue to modify these routines to use the recirq data collection idioms and `recirq.save` and `recirq.load`

recirq/otoc/utils.py
```diff
@@ -0,0 +1,253 @@
...
+
+    Returns:
+        The probabilities of all possible bit-strings for subsys_qubits,
+        ordered according to the big-endian notation.
+    """
+    num_qubits = len(subsys_qubits)
+    indices = [all_qubits.index(q) for q in subsys_qubits]
+    bits_red = np.asarray(bits[:, indices], dtype=int)
+    for i in range(num_qubits):
+        bits_red[:, i] *= 2 ** (num_qubits - i - 1)
```
mpharrigan, line 115: what are you doing here? cant you just do `np.bincount(bits_red)` and be done with it?

recirq/otoc/utils.py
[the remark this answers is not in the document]
mpharrigan: Yeah, we should use `PhasedFSimGate` as the implementation for this function or instead of this function. If you're not going to do it in this PR, please open an issue

recirq/otoc/utils.py
```diff
@@ -0,0 +1,253 @@
...
+    g_f.append(cirq.Z(q_0) ** (-(fsim_angles['delta_minus_diag'] -
+                                 fsim_angles['delta_minus_off_diag']) /
+                               np.pi / 4.0))
+    g_f.append(cirq.Z(q_1) ** ((fsim_angles['delta_minus_diag'] -
+                                fsim_angles['delta_minus_off_diag']) /
+                               np.pi / 4.0))
+    return g_f
+
+
+def cz_to_sqrt_iswap(qubit_0: cirq.GridQubit,
```
mpharrigan, line 213: Is this in Cirq or `cirq.google`? It probably should be

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+            single-qubit gates in each cycle. Otherwise, it will be applied
+            after the random single-qubit gates.
+        z_only: Whether the single-qubit gates are to be random \pi/2
+            rotations around axes on the equatorial plane of the Bloch sphere
+            (z_only = False), or random rotations around the z-axis (z_only =
+            True). In the former case, the axes of rotations will be chosen
+            randomly from 8 evenly spaced axes ($\pi/4$, $\pi/2$ ... $7\pi/4$
+            radians from the x-axis). In the latter case, the angles of
+            rotation will be any random value between $-\pi$ and $\pi$.
+        ancilla: If specified, an additional qubit will be included in the
```
mpharrigan, line 152: is this actually useful? Is the ancilla included in the analysis/fitting?

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+            lightcone are either removed or replaced with a spin-echo pulse.
+            Single-qubit gates outside the second lightcone, if specified,
+            are always removed.
+        echo_indices: An array with the same dimension as sq_rand_nums and
+            random integer values of 1, 2, 3 or 4. They specify the spin-echo
+            pulses applied to qubits outside the first lightcone, which can
+            be +/-X or +/-Y gates.
+
+    Returns:
+        A list of random circuits, each containing a specified number of cycles.
```
mpharrigan, line 170: document the second return value

recirq/otoc/parallel_xeb.py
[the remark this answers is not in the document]
mpharrigan: This exists as `cirq.experiments._default_interaction_sequence`.

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+        fsim_angles: Dict[str, float],
+        interaction_sequence: Optional[List[
+            Set[Tuple[Tuple[float, float], Tuple[float, float]]]]] = None,
+        gate_to_fit: str = 'iswap',
+        num_restarts: int = 3,
+        num_points: int = 8,
+        plot_individual_traces: bool = False,
+        plot_histograms: bool = False,
+        save_directory: Optional[str] = None
+) -> Tuple[Dict[Tuple[Tuple[int, int], Tuple[int, int]], cirq.Circuit],
```
mpharrigan, line 226: Make a dataclass to encapsulate the return value. This is too complex otherwise

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+            max_cycles = num_cycle_range[num_p - 1]
+
+            p_sim = [np.zeros((num_trials, 4)) for _ in
+                     range(max_cycles)]
+            for i in range(num_trials):
+                unitary = np.identity(4, dtype=complex)
+                for j in range(max_cycles):
+                    mat_0 = _rot_mats[sq_gates[(q0, q1)][i][0, j]]
+                    mat_1 = _rot_mats[sq_gates[(q0, q1)][i][1, j]]
+                    unitary = np.kron(mat_0, mat_1).dot(unitary)
```
mpharrigan, line 323: is this implementing a full statevector simulator within this closure?? Consider using `cirq.Circuit` and the built in cirq simulator functionality.

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+            err, x_vals, y_vals = pauli_error_fit(
+                np.asarray(num_cycle_range)[0:num_p],
+                np.asarray(fidelities), add_offset=False)
+
+            return err, fidelities, x_vals, y_vals, cost
+
+        def cost_function(angle_shifts: np.ndarray) -> float:
+            if gate_to_fit == 'sqrt-iswap':
+                full_shifts = np.zeros(5, dtype=float)
+                full_shifts[1:4] = angle_shifts
```
mpharrigan, line 343: lots of magic indices in this function. Add comments explaining or (better yet) use something more self-documenting like a dictionary

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+        corrected_angles['delta_plus'] *= -1.0
+        corrected_angles['delta_minus_off_diag'] *= -1.0
+        corrected_angles['delta_minus_diag'] *= -1.0
+        corrected_angles['theta'] = fsim_angles['theta']
+        corrected_angles['phi'] = fsim_angles['phi']
+        gate_list_corrected = generic_fsim_gate(corrected_angles, (q_0, q_1))
+        circuit_corrected = cirq.Circuit(gate_list_corrected)
+        correction_gates[(q0, q1)] = circuit_corrected
+
+        fig = plt.figure()
```
mpharrigan, line 438: separate plotting into its own function and have the user call it explicitly with the results of this function (if they desire)

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+            for j in range(len(num_cycle_range)):
+                p_data[j][i, :] = bits_to_probabilities(
+                    all_qubits, [q_index], measured_bits[i][j])
+
+        p_sim = [np.zeros((num_trials, 2)) for _ in
+                 range(max(num_cycle_range))]
+        for i in range(num_trials):
+            unitary = np.identity(2, dtype=complex)
+            for j in range(max(num_cycle_range)):
+                unitary = _rot_mats[scrambling_gates[i][k, j]].dot(unitary)
```
mpharrigan, line 554: Don't write your own simulator

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+                    p_sim[idx][i, :] = np.abs(unitary[:, 0]) ** 2
+
+        fidelities = [_alpha_least_square(p_sim[i], p_data[i]) for i in
+                      range(len(num_cycle_range))]
+
+        err, x_fitted, y_fitted = pauli_error_fit(
+            np.asarray(num_cycle_range), np.asarray(fidelities), num_qubits=1)
+        final_errors[q_index] = err
+
+        fig = plt.figure()
```
mpharrigan, line 566: Separate plotting into its own function that the user can call with the results of this function

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+def single_qubit_xeb_fidelities(
+        all_qubits: List[Tuple[int, int]],
+        num_cycle_range: Sequence[int],
+        measured_bits: List[List[np.ndarray]],
+        scrambling_gates: List[np.ndarray],
+        plot_individual_traces: bool = False,
+        plot_histograms: bool = False,
+        save_directory: Optional[str] = None
+) -> Dict[Tuple[int, int], float]:
+    """Computes single-qubit gate fidelities from parallel-XEB data.
```
mpharrigan, line 508: how does this work? Is the user supposed to run parallel single-qubit circuits?

recirq/otoc/parallel_xeb.py
```diff
@@ -0,0 +1,725 @@
...
+        denominator += delta_h ** 2
+    return nominator / denominator
+
+
+def _speckle_purity(probs_data: np.ndarray) -> float:
+    d = 4
+    return np.var(probs_data) * d ** 2 * (d + 1) / (d - 1)
+
+
+def _pairwise_xeb_probabilities(
```
mpharrigan, line 678: docstring for this function?

mpharrigan:
This also needs tests

recirq/otoc/utils.py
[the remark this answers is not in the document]
XiaoMiQC: I opened an issue on updating Cirq to allow usage of this gate.

recirq/otoc/parallel_xeb.py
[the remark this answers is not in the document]
XiaoMiQC: That's right (I actually copied the function here into Cirq...). This is now deleted from ReCirq.

recirq/otoc/otoc_example.py
XiaoMiQC: Done

recirq/otoc/otoc_example.py
XiaoMiQC: Fixed

recirq/otoc/otoc_example.py
XiaoMiQC: Will do

recirq/otoc/otoc_example.py
XiaoMiQC: Changed.

recirq/otoc/otoc_example.py
XiaoMiQC: Yes, it's just an easy way to add the 4 circuits related to measuring normalization at the beginning of `circuits_ic`. `build_otoc_circuits` always returns 16 circuits, the first 4 being the normalization circuits, and the next 12 being the circuits with the butterfly operator (X, Y, or Z) applied. Since the normalization curve is the same for any butterfly qubit, it only needs to be measured once. I therefore just added it to `circuits_ic` when the butterfly qubit is `qubits[1]`

recirq/otoc/otoc_example.py
XiaoMiQC: Somehow it still complains when I use the new sampler as defined above. Error message: Number of circuits and sweeps must match

recirq/otoc/otoc_example.py
XiaoMiQC: Done. As mentioned above, `params_list` still needs to be set in order for the function to run.

recirq/otoc/otoc_example.py
XiaoMiQC: Good catch. Indentation is wrong here. Fixed.

recirq/otoc/otoc_circuits.py
XiaoMiQC: It represents the two-qubit gates after parallel XEB correction for each qubit pair. It was called "file" because I usually save them... I changed the name to _GATE_CORRECTIONS to be more transparent and also documented this type.

recirq/otoc/otoc_circuits.py
XiaoMiQC: Good suggestion! Done.

recirq/otoc/otoc_circuits.py
XiaoMiQC: Done.

recirq/otoc/otoc_circuits.py
XiaoMiQC: Fixed.

mpharrigan:
hi @XiaoMiQC , I see you've responded to some of the comments but you haven't actually pushed any new changes. Did you forget to push?

recirq/otoc/otoc_circuits.py
XiaoMiQC: Done.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Done.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Renamed.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Renamed.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: This is now fixed in `build_xeb_circuits` (or rather, `_random_rotations`) where `cirq.value.parse_random_state(rand_seed)` is used

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: The second return value is now documented. I also renamed these variables to be more intuitive.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Done.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Okay, wrapped these in a data class. The _0 are sloppy names from previous versions and they are now deleted.

recirq/otoc/parallel_xeb_example.py
XiaoMiQC: Done, changed to current directory.

recirq/otoc/utils.py
XiaoMiQC: Yeah I agree that pickle seems to be fickle (no pun intended) with version compatibility. I will open an issue.

recirq/otoc/utils.py
XiaoMiQC: So this is computing the probability distribution from the bit-string in `bits_red`. It first turns each bit-string in `bits_red` into a base-10 number, then counts the number of occurrences for any base-10 number from 0 to 2**num_qubits.

Feeding bits_red directly to np.bincount would cause an error since bits_red is a 2D array (each row is a bit-string).

recirq/otoc/utils.py
XiaoMiQC: Not sure where it is. I simplified this function and added the corrections in the XEB function itself.

recirq/otoc/parallel_xeb.py
XiaoMiQC: The ancilla is not really used in parallel XEB. It is added here for convenience, because the same circuit generation function is used in `build_otoc_circuits()`. This saves a lot of trouble in the OTOC circuit generation code.

recirq/otoc/parallel_xeb.py
XiaoMiQC: Done

recirq/otoc/parallel_xeb.py
XiaoMiQC: Done

recirq/otoc/parallel_xeb.py
XiaoMiQC: Yeah, I tried that in the beginning. It was significantly slower than simply multiplying matrices here. Cirq simulator appears to only be faster when the number of qubits go up to 8 or so. Maybe the simulator has changed since then (~1 year ago), but I don't see much harm keeping this the way it is. It's just two qubits after all.

recirq/otoc/parallel_xeb.py
XiaoMiQC: Comments added to explain these indices.

recirq/otoc/parallel_xeb.py
XiaoMiQC: Done. Plotting is now a separate function call.

recirq/otoc/parallel_xeb.py
XiaoMiQC: See the comments above. Cirq simulation seems to be significantly slower for 2-qubit circuits which was the reason for writing these as simple matrix multiplications.

recirq/otoc/parallel_xeb.py
XiaoMiQC: Done

recirq/otoc/parallel_xeb.py
XiaoMiQC: This is historical. I deleted the single-qubit XEB code. It was occasionally used as a sanity check on the single-qubit gates, but hasn't been particularly useful for the OTOC experiment,

recirq/otoc/parallel_xeb.py
XiaoMiQC: Added docstring for this function,

recirq/otoc/otoc_example.py
```diff
@@ -77,33 +90,28 @@ def main():
...
             stats = int(2000 + 10000 * (c / max(cycles)) ** 3)
-            sweep_params = [{} for _ in range(len(circuits_ic))]
-            job = engine.run_batch(
-                programs=circuits_ic,
-                params_list=sweep_params,
-                repetitions=stats,
-                processor_ids=[processor_name],
-                gate_set=cirq.google.FSIM_GATESET)
+            params = [{} for _ in range(len(circuits_ic))]
+            job = sampler.run_batch(programs=circuits_ic, params_list=params, repetitions=stats)
```
mpharrigan, line 96: What happens if you omit `params_list`? IIRC this was broken at some point but I thought it was fixed so you don't have to supply this argument if you aren't using params.

recirq/otoc/otoc_example.py
```diff
@@ -77,33 +90,28 @@ def main():
...
             for d in range(num_qubits):
-                p = np.mean(job.results()[4 * d].measurements['z'])
-                p -= np.mean(job.results()[4 * d + 1].measurements['z'])
-                p -= np.mean(job.results()[4 * d + 2].measurements['z'])
-                p += np.mean(job.results()[4 * d + 3].measurements['z'])
+                p = np.mean(job[4 * d][0].measurements["z"])
+                p -= np.mean(job[4 * d + 1][0].measurements["z"])
+                p -= np.mean(job[4 * d + 2][0].measurements["z"])
+                p += np.mean(job[4 * d + 3][0].measurements["z"])
```
mpharrigan, line 101: These `[0]`. What are they? Is this from the "params" argument? If you omit `params_list` can you get rid of this `[0]`?

recirq/otoc/parallel_xeb.py
```diff
@@ -13,175 +13,219 @@
...
+]
 
 
-def default_interaction_sequence(
-        qubits: Sequence[cirq.GridQubit]
-) -> List[Set[Tuple[cirq.GridQubit, cirq.GridQubit]]]:
-    """Group qubits on a grid into sets of pairs.
+@dataclass
+class XEBData:
+    """Class for storing the cycle-dependent fidelities and purities of an XEB experiment."""
```
mpharrigan, line 55: Why are these all tuples of two ndarrays? Can you add a little note to the docstring?

recirq/otoc/parallel_xeb.py
```diff
@@ -13,175 +13,219 @@
...
-    After the sets are built as above, any empty set is ejected.:
+@dataclass
+class ParallelXEBResults:
+    """Class for storing results of a parallel-XEB experiment."""
 
-                 Cycle 1:                            Cycle 2:
-        q00 ── q01    q02 ── q03            q00    q01    q02    q03
-                                             |      |      |      |
-        q10 ── q11    q12 ── q13            q10    q11    q12    q13
+    fitted_gates: Dict[Tuple[Tuple[int, int], Tuple[int, int]], cirq.Circuit]
```
mpharrigan, line 69: Is `Tuple[Tuple[int, int], Tuple[int, int]]` a pair of grid qubits? Can we use `cirq.GridQubit` https://testing.googleblog.com/2017/11/obsessed-with-primitives.html

recirq/otoc/parallel_xeb.py
```diff
@@ -671,55 +589,94 @@ def _alpha_least_square(probs_exp: np.ndarray, probs_data: np.ndarray
...
-           Dict[Tuple[Tuple[int, int], Tuple[int, int]], List[np.ndarray]]]:
+    all_qubits: List[Tuple[int, int]],
+    num_cycle_range: Sequence[int],
+    measured_bits: List[List[List[np.ndarray]]],
+    scrambling_gates: List[List[np.ndarray]],
+    interaction_sequence: Optional[
+        List[Set[Tuple[Tuple[float, float], Tuple[float, float]]]]
+    ] = None,
+) -> Tuple[
+    Dict[Tuple[Tuple[int, int], Tuple[int, int]], List[np.ndarray]],
```
mpharrigan, line 606: use gridqubits instead of tuple tuple int int tuple int int

recirq/otoc/utils.py
mpharrigan: Yeah. I think instead of

```python
bits = list(np.sum(bits_red, axis=1))
p_reduced = [bits.count(i) for i in range(2 ** num_qubits)]
```

You could use bincount in the second line (and not turn it to a list in the first line). This will likely be faster, but may not make a difference if this function isn't in a hot path

mpharrigan (commented):
Thanks for going through my comments! Just a couple small things I noticed in this review. Please let me know what you think

recirq/otoc/otoc_example.py
mpharrigan: hmmm are you using the latest cirq? This sounds like a bug. Can you open a Cirq issue?

recirq/otoc/otoc_example.py
XiaoMiQC: This is the error message: "ValueError: Number of circuits and sweeps must match"

It is a bit odd since if I set the sampler to be `cirq.Simulator() `instead of a processor, it doesn't raise this error message even if I omit `param_list`. This also appears to be a Cirq version issue. I'm using Cirq 0.9.1 as listed in the requirements.txt. This problem doesn't exist if it is upgraded to 0.10.

I will raise an issue here to call for Cirq upgrade.

recirq/otoc/otoc_example.py
XiaoMiQC: I don't know what they are. I have to specify them even if I omit params_list with Cirq v0.10. In the `run_batch method`, the return type is specified to be `List[List['cirq.Result']]`, so it is definitely a nested list for some reason.

recirq/otoc/parallel_xeb.py
XiaoMiQC: They are just the X (cycle) and Y (XEB fidelity or speckle purity) values of the decay curves and their fits. I added an explanation here.

recirq/otoc/parallel_xeb.py
XiaoMiQC: Yeah that's actually what I did in the past, until we run into issues with saving them and loading them back in with pickle formats... The code won't recognize the keys when Cirq gets updated. It might be an issue with pickle and maybe some other saving formats won't make it more robust. Perhaps we can leave this for later when we have a better way to save generic data?

recirq/otoc/parallel_xeb.py
XiaoMiQC: Same as above. The code sometimes doesn't recognize the keys if the results are saved/reloaded while Cirq gets updated in-between.

recirq/otoc/otoc_example.py
mpharrigan: ok cool, that makes sense

recirq/otoc/otoc_example.py
mpharrigan: 🤷‍♂️

recirq/otoc/parallel_xeb.py
mpharrigan: Oh my goodness! This is https://github.com/quantumlib/Cirq/issues/3777 which was mania to figure out. You should have opened a bug report and saved me headache :)

XiaoMiQC: Ah, I might have mentioned this somewhere and you (or maybe someone else) told me I should not save data with pickle =) So i thought this hashing thing was intentional and didn't press it further.

mpharrigan (approved):

Issue #61, opened 2020-08-19
[Requirement Package Version] cirq=0.8.2 is not compatible with 


JiahaoYao:
Hi @mpharrigan, I met with the install error for the package. Here are the description of the error and my approach to getting around. 

```
Installed /home/ABC/anaconda3/envs/recirq/lib/python3.7/site-packages/recirq-0.1.dev0-py3.7.egg
Processing dependencies for recirq==0.1.dev0
Searching for cirq==0.8.2
Reading https://pypi.org/simple/cirq/
Downloading https://files.pythonhosted.org/packages/1e/5b/6f8cb54ea8c0041ad9c8e4ece07cb5ca9eb1c29de68e68795b4a40d90cc6/cirq-0.8.2-py3-none-any.whl#sha256=84c2ace7f6cb96b21e9265ea31a25c2fbc044ee79da30beb83fe3865e2cc05f8
Best match: cirq 0.8.2
Processing cirq-0.8.2-py3-none-any.whl
Installing cirq-0.8.2-py3-none-any.whl to /home/ABC/anaconda3/envs/recirq/lib/python3.7/site-packages
Adding cirq 0.8.2 to easy-install.pth file

Installed /home/ABC/anaconda3/envs/recirq/lib/python3.7/site-packages/cirq-0.8.2-py3.7.egg
error: cirq 0.8.2 is installed but cirq==0.8 is required by {'pytket-cirq'}
```

I think this is related to the [requirement.txt](https://github.com/quantumlib/ReCirq/blob/master/requirements.txt#L1). 

If it is changed to `cirq=0.8.0`, there is no such issue.

mpharrigan:
It looks like this is because `pytket-cirq` (a 3rd party dependency) is pinned to version 0.8, but we actually applied a couple bug fixes (ironically around dependency management) that means we really want to be using 0.8.2. In an ideal world, the pyket folks would use a looser pin, i.e. `cirq~=0.8` so it wouldn't cause an unsatisfiable dependency.

I know it says error, but does it legitimately not work? As I understand it, `pip` is fairly forgiving and will proceed anyway

JiahaoYao:
yes, you are right! Thanks so much for your reply! Hope pyket folks can fix that!
And I am closing #61.

PR #99, opened 2020-11-19
Fix "guide" notebooks to run under colab

mpharrigan:
- fix a couple of broken links
 - stop using a couple of deprecated cirq functions
 - fixup the dev_tools script that converts these notebooks to recirq.readout_scan.tasks
 - **main**: Upload the guide sample data to figshare. Add utility function to fetch it so notebooks are self-contained

https://figshare.com/articles/dataset/Readout_Scan_Tutorial_Data/13262873


docs/guide/data_analysis.ipynb
```diff
@@ -93,7 +93,7 @@
         "try:\n",
         "    import recirq\n",
         "except ImportError:\n",
-        "    !pip install git+https://github.com/quantumlib/ReCirq"
+        "    !pip install --quiet git+https://github.com/quantumlib/ReCirq"
```
lamberta, line 96: Getting an install error when I run in Colab: https://colab.research.google.com/github/mpharrigan/ReCirq/blob/2020-11-docs-1/docs/guide/data_collection.ipynb

```
ERROR: pytket 0.6.1 has requirement sympy~=1.5, but you'll have sympy 1.1.1 which is incompatible.
ERROR: pytket-cirq 0.5.0 has requirement cirq~=0.8.0, but you'll have cirq 0.9.1 which is incompatible.
```

The Colab environment might default to a different version than you need. Or maybe pin a version when pip installing?

mpharrigan: Yeah, it says `ERROR` but it should still work. The message is a result of overzealous pinning in cirq *and* pytket

mpharrigan: xref #61 and https://github.com/CQCL/pytket/issues/22 -- which I fixed but looks like we need to bump again. It doesn't help that pytket is closed source so they have to make the change

docs/guide/data_analysis.ipynb
```diff
@@ -127,7 +127,20 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "75m62Mlkphm2"
+      },
+      "outputs": [],
+      "source": [
+        "recirq.fetch_guide_data_collection_data()"
```
lamberta, line 143: Got this error in Colab:

```
AttributeError: module 'recirq' has no attribute 'fetch_guide_data_collection_data'
```

mpharrigan: That function is added in this PR, so you can't really test it until this is merged

lamberta: Gotcha. LGTM and we can test once it lands. Thanks

lamberta (approved):

mpharrigan:
Just verified it runs in colab after merge

PR #103, opened 2020-11-30
Format quantum_chess_rest_api.ipynb

lamberta:
Add cell to disable local webserver in docs infra.

### DIFF
--- a/docs/_toc.yaml
+++ b/docs/_toc.yaml
@@ -35,9 +35,9 @@ toc:
   path: /cirq/experiments/quantum_chess
 - title: "Concepts"
   path: /cirq/experiments/quantum_chess/concepts
-  title: "Quantum Chess REST API"
+- title: "Quantum Chess REST API"
   path: /cirq/experiments/quantum_chess/quantum_chess_rest_api
-  title: "Quantum Chess client"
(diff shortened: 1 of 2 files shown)


mpharrigan (approved):
lgtm

do we eventually want to put a `try: import` guard around the flask pip installs? If we're not executing the web server, should we skip executing the flask-install and flask server definition cells as well? To save time when executing the notebook in the CI

mpharrigan:
Debugging build failure in https://github.com/quantumlib/ReCirq/issues/104

lamberta:
The server is still run in Colab, just not when converting to Markdown for the site now.

Probably need to keep the pip installs since we still need it to define the routes for the webserver. But, as far as GItHub Actions CI goes, may want to guard that final `app.run()` if it's causing things to hang. Alternatively, could make that a code formatted section within a text cell, but that would not execute anywhere.

dstrain115 (approved):

mpharrigan:
@lamberta can you merge in master to pick up https://github.com/quantumlib/ReCirq/pull/105 which will fix the CI

lamberta:
@mpharrigan Rebased and looks like it worked. Thanks!

Issue #97, opened 2020-11-17
Non-deterministic test failure in quantum chess


rmlarose:
Noticed a failure in [test_undo_entangled_measurement](https://github.com/quantumlib/ReCirq/blob/master/recirq/quantum_chess/quantum_board_test.py#L873) on PR #96 which is unrelated to quantum chess. It doesn't seem to fail often but I can reproduce it after running several times.

```
=================================== FAILURES ===================================
_______________________ test_undo_entangled_measurement ________________________

    def test_undo_entangled_measurement():
        b = qb.CirqBoard(u.squares_to_bitboard(['a2','b1','c2','d1']))
        assert b.perform_moves(
            'b1a3c3:SPLIT_JUMP:BASIC',
            'c2c4:PAWN_TWO_STEP:BASIC'
        )
        probs = b.get_probability_distribution(1000)
        assert_prob_about(probs, qb.square_to_bit('a3'), 0.5)
        assert_prob_about(probs, qb.square_to_bit('c2'), 0.5)
        assert_prob_about(probs, qb.square_to_bit('c3'), 0.5)
        assert_prob_about(probs, qb.square_to_bit('c4'), 0.5)
        b.perform_moves( 'd1c2:JUMP:EXCLUDED')
        assert b.undo_last_move()
        print(b)
        probs = b.get_probability_distribution(1000)
>       assert_prob_about(probs, qb.square_to_bit('a3'), 0.5)

recirq/quantum_chess/quantum_board_test.py:888: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

probs = [0.0, 0.0, 0.0, 1.0, 0.0, 0.0, ...], that = 16, expected = 0.5
atol = 0.04

    def assert_prob_about(probs, that, expected, atol=0.04):
        """Checks that the probability is within atol of the expected value."""
        assert probs[that] > expected - atol
>       assert probs[that] < expected + atol
E       assert 0.54 < (0.5 + 0.04)

recirq/quantum_chess/test_utils.py:83: AssertionError
----------------------------- Captured stdout call -----------------------------
 +----------------------------------+
8|  .   .   .   .   .   .   .   .   |
7|  .   .   .   .   .   .   .   .   |
6|  .   .   .   .   .   .   .   .   |
5|  .   .   .   .   .   .   .   .   |
[...]

PR #106, opened 2020-12-02
Fix failing Fermi-Hubbard experiment example

rmlarose:
The final portion of the notebook demonstrates how to use the QCS. To fix the authentication error I defaulted to a simulator:

```python
try:
    engine_sampler = cirq.google.get_engine_sampler(
        processor_id="rainbow", gate_set_name="sqrt_iswap"
    )
except OSError:
    # Use the simulator as a backup.
    engine_sampler = cirq.Simulator()
```

 Other options are possible, let me know what you think.


mpharrigan:
It's strange to expect `OSError` to signal lack of authentication. In practice, this is because the actual line that fails is looking for the google cloud environment variable. What if the conditional was instead 

```python
if 'GOOGLE_CLOUD_PROJECT' in os.environ:
  blah
else:
  simulator
```

mpharrigan (approved):
looks good. one suggestion in the thread for your consideration

rmlarose:
> ```python
> GOOGLE_CLOUD_PROJECT
> ```

I like it!

rmlarose:
Unfortunate random failure due to #97. Changed files should be ok to merge still.

mrwojtek (approved):

PR #107, opened 2020-12-02
Fix failing fermi hubbard publication results notebook

rmlarose:
Build was failing due to referencing non-existent data. Added cell to download data (if not present) to fix.

Also added copyright/frontmatter (links), but there could be some more text explaining the plots. But for now this at least should fix the failing test.


lamberta:
Thanks! Verified it runs in Colab: https://colab.research.google.com/github/rmlarose/ReCirq/blob/qafh/docs/fermi_hubbard/publication_results.ipynb

Can you please format the notebook and clear the outputs so we can test it in the docs infra?

```
$ python3 -m pip install -U --user git+https://github.com/tensorflow/docs

$ python3 -m tensorflow_docs.tools.nbfmt --indent=1 --remove_outputs notebook.ipynb
```

mpharrigan:
GitHub refuses to render the diff because it is too large (?) I think @lamberta's suggestion is a must!

Do you have any appetite for factoring out the download logic into a helper function like https://github.com/quantumlib/ReCirq/blob/master/recirq/documentation_utils.py#L56

bash for loops in jupyter magic cells makes me queasy

rmlarose:
Oops, forget to format! Should be ok now.

@mpharrigan The suggestion for a download helper sounds great to me. I could do it in this PR or a separate one if this needs to get merged ASAP - what do you think? I lean towards merging now (pending review) since there is still other QA to do in this tutorial, but either are fine for me.

mpharrigan:
I think we were supposed to have fixed this by yesterday so maybe a follow-on PR is appropriate :) I've opened #108

mpharrigan (approved):

mrwojtek (approved):
This is cool.

mpharrigan:
xref #83

PR #117, opened 2020-12-14
Add helper function for downloading Fermi-Hubbard publication data

### LINKED ISSUE #108: Helper function for downloading fermi hubbard experimental data
See https://github.com/quantumlib/ReCirq/pull/107#issuecomment-737513496

rmlarose:
Fixes #108.

### DIFF
--- a/recirq/fermi_hubbard/publication.py
+++ b/recirq/fermi_hubbard/publication.py
@@ -11,11 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Data specific to experiment published in arXiv:2009.XXXXX."""
+"""Data specific to experiment published in arXiv:2010.07965."""
 
-from typing import Callable, Tuple
-
+from io import BytesIO
 from copy import deepcopy
+import os
+from typing import Callable, List, Optional, Tuple
+from urllib.request import urlopen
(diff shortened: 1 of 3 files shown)


recirq/fermi_hubbard/publication.py
```diff
@@ -209,3 +213,50 @@ def rainbow23_layouts(sites_count: int = 8) -> Tuple[ZigZagLayout]:
                  for flip in (False, True)
                  for exchange in (False, True)
                  for reverse in (False, True))
+
+
+def fetch_publication_data(
```
mpharrigan, line 218: The other function is in `recirq.documentation_utils`. I don't know if that's the exact correct place to put it, but I think there's some value to having all the example-dataset functionality outside of the projects so you can find all of them more easily in one place. What do you think?

rmlarose: I guess there are two options which make sense:

1. Put download functions in the same place (`documentation_utils.get_qaoa_data` and `documentation_utils.get_fermi_hubbard_data`).
2. Put download functions in respective modules (`qaoa.get_publication_data` and `recirq.get_publication_data`).

I'm not too partial to either so I'm good to go with your intuition here. Maybe (2) is slightly more natural to me since I would probably look there before `documentation_utils` if I were guessing. This could be improved by renaming `documentation_utils` or putting both functions in a new module though.

mpharrigan (approved):
No sense letting the discussion hold this up

Issue #104, opened 2020-12-01
PyTket dependency is resolving badly


mpharrigan:
The build is failing.

https://github.com/quantumlib/ReCirq/runs/1475828878?check_suite_focus=true#step:6:47

Need to check if they released a new version and why it's not backwards compatible. Pin or fix as necessary

mpharrigan:
I can't reproduce this issue locally with the latest versions of pytket and pytket-cirq

mpharrigan:
There might be something funky with the (new?) pip dependency resolver where it is installing pytket 0.6.1 (latest) but pytket-cirq 0.3 (old)

mpharrigan:
pytket-cirq 0.5 has the following `install_requires`

```python
    install_requires=["pytket ~= 0.6", "cirq ~= 0.8.0"],
```

mpharrigan:
If I pin all three packages, it warns that pip will break when they flip on the "new resolver".

mpharrigan:
The pypa people -- with a straight face -- propose forking dependencies to work around these types of issues! https://pip.pypa.io/en/latest/user_guide/#loosen-the-requirements-of-your-dependencies

PR #118, opened 2020-12-15
Bump pytket-cirq dependency to resolve conflict

rmlarose:
There is a dependency conflict when installing ReCirq which causes the latest version of pip (20.3.3) to fail (see below). This PR bumps the `pytket-cirq` dependency to v0.5.1 which uses Cirq v0.9 and resolves the conflict

```
pip install git+https://github.com/quantumlib/recirq

...

ERROR: Cannot install recirq and recirq==0.1.dev0 because these package versions have conflicting dependencies.

The conflict is caused by:
    recirq 0.1.dev0 depends on cirq==0.9.1
    pytket-cirq 0.5.0 depends on cirq~=0.8.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ scikit-learn
 # qaoa
 networkx
 pytket==0.6.1
-pytket-cirq==0.5
+pytket-cirq==0.5.1
 
 # quantum chess, only needed for tests
 scipy


mpharrigan:
xref #104

mpharrigan:
It would be my preference to just revert https://github.com/quantumlib/ReCirq/pull/105/files

Since we use `requirements.txt` contents for our `install_requires` I don't like being too picky about pinning, or else you get issues like this!

rmlarose:
Thanks @mpharrigan! I will close this in favor of #119 which I agree is a better solution.

PR #120, opened 2021-01-07
Use tket 2D qubit indices with new register name

ss2165:
Previously it seems 2D registers (as the equivalent of `GridQubit`) were used with "q" as the register name, which lead to failure as "q" is a reserved name in tket for the default 1D register (this should be made more clear in the tket docs/error message). I have changed the name to "grid", and removed the workaround used to map 1D indices to 2D qubits.

### DIFF
--- a/docs/qaoa/routing_with_tket.ipynb
+++ b/docs/qaoa/routing_with_tket.ipynb
@@ -225,28 +225,6 @@
     {
       "cell_type": "markdown",
       "metadata": {
-        "id": "UZMHX0TfLzh_"
-      },
-      "source": [
-        "### Mapping to device *indices*\n",
-        "We'll keep a set of secret indices that number device qubits contiguously from zero instead of `(row, col)`"
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": null,
-      "metadata": {
-        "id": "qWak5p0QLziB"
-      },
-      "outputs": [],
-      "source": [
-        "index_to_qubit = sorted(device.qubit_set())\n",
-        "qubit_to_index = {q: i for i, q in enumerate(index_to_qubit)}"
-      ]
-    },
-    {
-      "cell_type": "markdown",
-      "metadata": {
         "id": "G79yrP3_LziE"
       },
       "source": [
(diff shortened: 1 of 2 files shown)


mpharrigan:
Thanks for the patch. I will take a closer look shortly. I'm trying to remember what motivated this originally. It must have been the "q" undocumented failure. 

Do you want to also patch `recirq.qaoa` code to use 2d registers as well?

ss2165:
Thanks, I've replicated the patch in `recirq.qaoa.placement` now, is there anywhere else I should to edit?

docs/qaoa/routing_with_tket.ipynb
```diff
@@ -401,12 +367,11 @@
       },
       "outputs": [],
       "source": [
-        "def tk_to_i(tk):\n",
-        "    i = tk.index\n",
-        "    assert len(i) == 1, i\n",
-        "    return i[0]\n",
+        "def tk_to_cirq_qubit(tk):\n",
+        "    ind = tk.index\n",
+        "    return cirq.LineQubit(ind[0]) if len(ind) == 1 else cirq.GridQid(*ind, dimension=len(ind))\n",
```
mpharrigan, line 372: the `dimension` argument being equal to `len(ind)` is a coincidence and should be omitted. This is to support qudits. Just use `cirq.GridQubit`.

recirq/qaoa/placement.py
```diff
@@ -37,33 +38,34 @@ def calibration_data_to_graph(calib_dict: Dict) -> nx.Graph:
...
-    return i[0]
+def tk_to_cirq_qubit(tk: Qubit):
+    """Convert a tket Qubit to either a LineQubit or GridQubit.
+
+    """
+    ind = tk.index
+    return (
+        cirq.LineQubit(ind[0])
+        if len(ind) == 1
+        else cirq.GridQid(*ind, dimension=len(ind))
```
mpharrigan, line 67: ditto, use `cirq.GridQubit` and omit `dimension`.

mpharrigan (changes requested):
Thanks again! This really cleans things up. Left a couple comments

ss2165:
Thank you for the comments, I've made the suggested changes. I must confess I actually misread `GridQid` as `GridQubit`...

mpharrigan (approved):
Thanks!

PR #121, opened 2021-01-07
[docs] Fix hfvqe molecular data notebook

rmlarose:
This notebook installs openfermionpyscf v0.5.0 which is incompatible with recirq (requires openfermion 0.11.0). This small patch pins openfermionpyscf to v0.4.0 to fix the docs build.


mpharrigan:
Yes, let's do this for now but open an issue to migrate to latest openfermion

mpharrigan (approved):

PR #124, opened 2021-01-11
Add notebook formatting and CI check

### LINKED ISSUE #56: Add nbfmt check to CI
See https://github.com/quantumlib/ReCirq/pull/52#issuecomment-672257227 and following #55

rmlarose:
Fixes #56.


dev_tools/nbformat
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Formats ipython notebooks with tensorflow-docs nbformat tool.
```
mpharrigan, line 4: Did you write this or take it from somewhere? 

If the former: ick bash. I guess it's ok here since you aren't doing looping (which is my rule of thumb for when a script definitely needs to be in a real programming langauge)

If the latter: note where you got it from so we can keep them in sync

mpharrigan (commented):
Oof that diff.

I'll need to download this and look at it in pycharm's differ (which handles the change in indent much more elegantly). Is it just changing the indent or are there additional changes? 

How hard would it be to maintain a list of grandfathered-in 2-indent notebooks so we can kick this can down the road some more?

dev_tools/nbformat
rmlarose: This is taken directly from Cirq: https://github.com/quantumlib/Cirq/blob/master/check/nbformat

In addition to indentation, metadata is also removed and probably a few other things - @lamberta may know more here.

Not sure why the diff is so large - may be because GH has trouble analyzing notebooks but I'll check. At least by quick glance the notebooks are very similar before/after formatting.

lamberta: Try passing `--indent=2` to nbfmt. The default is 4 but the quantum folks prefer 2

dev_tools/nbformat
```diff
@@ -0,0 +1,61 @@
...
+        exit 1
+    fi
+done
+
+# Get the working directory to the repo root.
+cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "$(git rev-parse --show-toplevel)"
+
+# Install a pinned version of tf-docs
+TF_DOCS_COMMIT=b3dc8a922d8bdc6e998a8ad4f4953359dc6e576a
```
lamberta, line 32: Pinning a commit makes sense, but we don't really do releases in tensorflow/docs so this commit is arbitrary.

It might be better to download the cirq `dev_tools/nbformat` script into your GitHub Actions VM so you're at least using the same thing and don't get too out of sync.

rmlarose:
Diff is about 1/2 as large with indent=2. (But note [Cirq sets indent=1](https://github.com/quantumlib/Cirq/blob/master/check/nbformat#L39).) 

> How hard would it be to maintain a list of grandfathered-in 2-indent notebooks so we can kick this can down the road some more?

I think it's fine to punt on this. This PR is the simplest fix for #56 using the tool already in Cirq, but it sounds like there's some things to decide on.

> It might be better to download the cirq dev_tools/nbformat script into your GitHub Actions VM so you're at least using the same thing and don't get too out of sync.

Great point. So would we want to just [grab this file](https://stackoverflow.com/questions/4604663/download-single-files-from-github) from Cirq in CI to make sure they're the same? I guess this works as long as the file location doesn't change.

lamberta:
Ah, you're right (and I misspoke): nbfmt indent [defaults to 2](https://github.com/tensorflow/docs/blob/master/tools/tensorflow_docs/tools/nbfmt/__main__.py#L48) and the quantum folks mentioned they prefer 1.

>  So would we want to just grab this file from Cirq in CI to make sure they're the same?

Curl'ing the file is probably easiest for a one-off. Though you can checkout the entire Cirq repo (shallow clone) next to this one using the GitHub Actions checkout module—if there's additional tooling in that repo you needed to access.

dev_tools/nbformat
mpharrigan: @lamberta I think the collab default is 2 but the jupyter default is 1 and we prefer 1

@rmlarose the diff is messed up because of the indentation of the json in the ipynb files. Currently recirq has some with 2 and some with 1

dev_tools/nbformat
mpharrigan: +1 downloading cirq's script

mpharrigan:
> How hard would it be to maintain a list of grandfathered-in 2-indent notebooks

If we use Cirq's script (which I think we should) then we can't do this (without a lot of headache) so I'm fine "biting the bullet" now. 

Like I said, pycharm seems to handle the change-in-indent diffs much better so I'll review the PR that way once it's finalized with indent=1 and using cirq's script

mpharrigan:
Another fun trick is that you can append `?w=1` to a github diff url to omit whitespace changes:

https://github.com/quantumlib/ReCirq/pull/124/files?w=1

mpharrigan:
@rmlarose what do you think about `wget`ting the script from Cirq?

rmlarose:
> @rmlarose what do you think about `wget`ting the script from Cirq?

I like it! I should be able to get to this by the end of the week, just haven't had time yet.

Only potential downside I see is a new contributor to ReCirq wondering how to format a notebook to pass CI. It can be figured out though, and I agree its a priority to be in sync with Cirq.

mpharrigan:
> Only potential downside I see is a new contributor to ReCirq wondering how to format a notebook to pass CI. It can be figured out though, and I agree its a priority to be in sync with Cirq.

Good point. Maybe we can have a simple shell script in this repo to optionally download and then execute the script

```bash
if [[ check/nbfmt doesn't exist ]]
  wget it from cirq
fi

check/nbfmt
```

which the CI can just call (as well as contributors). [and add the downloaded file to `.gitignore` so people don't commit it]

lamberta:
> Only potential downside I see is a new contributor to ReCirq wondering how to format a notebook to pass CI.

True. Depending how fancy you want to get, you can add a GitHub Action to automatically post a comment with instructions to the pull request, for example: https://github.com/tensorflow/docs-l10n/pull/474#issuecomment-761102578
But the simplest thing is to add a saved reply to your GitHub settings and use that if you're monitoring PRs. And if you have a open source contributing guide you can add instructions there (at least you'll have a consistent place to point).

We actually have a bot set up to automatically format new pull requests. It's nice but requires auth setup for GitHub with a Cloud Function. Probably not worth it for now

rmlarose:
> Good point. Maybe we can have a simple shell script in this repo to optionally download and then execute the script
> 
> ```shell
> if [[ check/nbfmt doesn't exist ]]
>   wget it from cirq
> fi
> 
> check/nbfmt
> ```
> 
> which the CI can just call (as well as contributors). [and add the downloaded file to `.gitignore` so people don't commit it]

Went with this method. 

- One can do `dev_tools/nbfmt [--apply]` locally which will grab the Cirq `nbformat` script (if it doesn't already exist) then run it. 
- `nbformat` added to gitignore.
- Back to large `--indent=1` diff, but on par with Cirq.

Works well I think, but I guess we're kind of crossing our fingers that no one moves or renames `nbformat` in Cirq.

mpharrigan:
> we're kind of crossing our fingers that no one moves or renames nbformat in Cirq.

This would break the CI but not any user-facing functionality, i.e. the least-bad breakage. I'll keep an eye on Cirq PRs with this in mind anyways though :)

.gitignore
```diff
@@ -39,3 +39,6 @@ docs/generated
 
 # Default pycharm virtual env
 .venv/
+
+# Notebook formatting script.
+nbformat
```
mpharrigan, line 44: but it downloads into `dev_tools/`, right? I guess this still works but may be overzealous. Probably doesn't matter too much

docs/quantum_chess/concepts.ipynb
```diff
@@ -1,2163 +1,2063 @@
...
-      "cell_type": "code",
-      "metadata": {
-        "id": "GZILs0lQpFOE",
-        "outputId": "3c4f305a-d627-4429-a565-174083cab0eb",
-        "colab": {
-          "base_uri": "https://localhost:8080/"
-        }
-      },
-      "source": [
-        "king_moves = cirq.Circuit(\n",
```
mpharrigan, line 434: is this deleting code?

mpharrigan: I guess it's moving it down lower

mpharrigan (approved):
Checked the diff offline. Looks good; quantum chess notebook is a little wonky. Why are we saving outputs there? And if we are, we should run it through "fresh" so the indices are contiguous from zero

rmlarose:
> quantum chess notebook is a little wonky. Why are we saving outputs there? And if we are, we should run it through "fresh" so the indices are contiguous from zero

Not sure. Maybe there was a good reason for saving output? The first cell starting at 97 is certainly not necessary - at least this doesn't show on the website or on colab though. 

See #135 which follows up on this. I can address any remaining comments from this PR there if desired.

.gitignore
rmlarose: Indeed downloads into `dev_tools`

PR #135, opened 2021-01-26
Better error message for nbfmt

rmlarose:
Follow up from #124. The `nbfmt` error message is the "Cirq error message" which says to run `check/nbformat` which doesn't exist in ReCirq. This PR updates the error message to reference the correct file, namely `dev_tools/nbfmt`. For example:

```bash
The following notebooks require formatting.
- docs/fermi_hubbard/experiment_example.ipynb
Notebooks are not formatted. Please run 'dev_tools/nbfmt --apply'
```

### DIFF
--- a/docs/fermi_hubbard/experiment_example.ipynb
+++ b/docs/fermi_hubbard/experiment_example.ipynb
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 4,
    "metadata": {
     "cellView": "form",
     "id": "906e07f6e562"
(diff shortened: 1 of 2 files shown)


docs/fermi_hubbard/experiment_example.ipynb
```diff
@@ -11,7 +11,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
```
mpharrigan, line 14: why is this change included?

mpharrigan (commented):
haha awesome

docs/fermi_hubbard/experiment_example.ipynb
rmlarose: Ah, leftover from intentional unformatting. Fixed now.

mpharrigan (approved):

PR #136, opened 2021-01-30
log time for sample_with_ancilla

madcpf:
Adds the elapsed time for each call of sample_with_ancilla into self.debug_log.

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import time
+from collections import defaultdict
 from typing import Dict, List, Optional, Tuple
 
 import cirq
@@ -94,7 +95,7 @@ class CirqBoard:
         # the move-history when undoing moves
         self.init_basis_state = basis_state
         self.clear_debug_log()
-        self.timing_stats = []
+        self.timing_stats = defaultdict(list)
         return self
 
     def clear_debug_log(self) -> None:
@@ -109,7 +110,7 @@ class CirqBoard:
             
     def clear_timing_stats(self) -> None:
         """Clears timing stats."""
-        self.timing_stats = []
+        self.timing_stats = defaultdict(list)
 
     def print_timing_stats(self, clear_stats: bool = False) -> None:
         """Prints timing stats. Clears timing stats if clears_stats is enabled."""
@@ -154,7 +155,7 @@ class CirqBoard:
         if t1 is None:
             t1 = time.perf_counter()
         self.debug_log += (f"{action} takes {t1 - t0:0.4f} seconds.\n")
-        self.timing_stats.append(t1 - t0)
+        self.timing_stats[action].append(t1 - t0)
     
     def sample_with_ancilla(self, num_samples: int
                             ) -> Tuple[List[int], List[Dict[str, int]]]:
@@ -279,7 +280,7 @@ class CirqBoard:
                         f'Discarded {error_count} from error mitigation '
                         f'{noise_count} from noise and '
                         f'{post_count} from post-selection\n')
-                    self.record_time("sample_with_ancilla", t0)
+                    self.record_time('sample_with_ancilla', t0)
                     return (rtn, ancilla)
         else:
             rtn = [self.state] * num_samples
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -94,6 +94,7 @@ class CirqBoard:
         # the move-history when undoing moves
         self.init_basis_state = basis_state
         self.clear_debug_log()
+        self.timing_stats = []
         return self
 
     def clear_debug_log(self) -> None:
@@ -259,6 +260,7 @@ class CirqBoard:
                         f'{post_count} from post-selection\n')
                     t1 = time.perf_counter()
                     self.debug_log += (f"sample_with_ancilla takes {t1 - t0:0.4f} seconds.\n")
+                    self.timing_stats.append(t1 - t0)
                     return (rtn, ancilla)
         else:
             rtn = [self.state] * num_samples
@@ -268,6 +270,7 @@ class CirqBoard:
             )
         t1 = time.perf_counter()
         self.debug_log += (f"sample_with_ancilla takes {t1 - t0:0.4f} seconds.\n")
+        self.timing_stats.append(t1 - t0)
         return (rtn, ancilla)
 
     def sample(self, num_samples: int) -> List[int]:
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -104,6 +104,14 @@ class CirqBoard:
         print(self.debug_log)
         if clear_log:
             self.clear_debug_log()
+            
+    def clear_timing_stats(self) -> None:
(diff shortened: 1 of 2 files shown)


madcpf:
@googlebot I signed it!

recirq/quantum_chess/quantum_board.py
```diff
@@ -255,13 +257,17 @@ def sample_with_ancilla(self, num_samples: int
                         f'Discarded {error_count} from error mitigation '
                         f'{noise_count} from noise and '
                         f'{post_count} from post-selection\n')
+                    t1 = time.perf_counter()
+                    self.debug_log += (f"sample_with_ancilla takes {t1 - t0:0.4f} seconds.\n")
```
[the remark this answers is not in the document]
madcpf: Thanks Doug. I've added self.timing_stats.

recirq/quantum_chess/quantum_board.py
```diff
@@ -255,13 +258,19 @@ def sample_with_ancilla(self, num_samples: int
...
+                    self.timing_stats.append(t1 - t0)
```
dstrain115, line 263: Should we separate this into a function, since it is called twice?  That will make it easier to keep both branches in sync.

recirq/quantum_chess/quantum_board.py
```diff
@@ -255,13 +258,19 @@ def sample_with_ancilla(self, num_samples: int
...
                     return (rtn, ancilla)
         else:
             rtn = [self.state] * num_samples
             self.debug_log += (
                 f'Discarded {error_count} from error mitigation '
                 f'{noise_count} from noise and {post_count} from post-selection\n'
             )
+        t1 = time.perf_counter()
+        self.debug_log += (f"sample_with_ancilla takes {t1 - t0:0.4f} seconds.\n")
+        self.timing_stats.append(t1 - t0)
```
dstrain115, line 273: Let's add a short test for this functionality as well.

recirq/quantum_chess/quantum_board.py
```diff
@@ -133,6 +135,10 @@ def undo_last_move(self) -> bool:
                 return False
         return True
 
+    def record_time(self, t0, t1) -> None:
```
dstrain115, line 138: You could have t1 be an optional argument and if None, set it to time.perf_counter()

madcpf: Thanks Doug for looking into this! The function record_time() is updated.

recirq/quantum_chess/quantum_board.py
madcpf: Thanks Doug. What is the right way to run tests? I've tried to add the following lines to quantum_board_test.py:

```
def test_record_time():
    b = qb.CirqBoard(u.squares_to_bitboard(['b8', 'b5', 'c7']))
    assert b.perform_moves(
        'b8a6c6:SPLIT_JUMP:BASIC',
        'b5a6:PAWN_CAPTURE:BASIC',
        'c6b8:JUMP:BASIC',
        'c7c5:PAWN_TWO_STEP:BASIC',
        'b5c6:PAWN_EP:BASIC',
    )
    b.print_timing_stats()

def main():
    test_record_time()

if __name__ == "__main__":
    main()
```

But it returns the error:
AttributeError: 'CirqBoard' object has no attribute 'print_timing_stats'

dstrain115: Don't add the main() function to the test.  You can just add test_record_time (or any function that begins with "test_" and then run "pytest path/to/quantum_board_test.py".  

I am not sure why you are getting the error.  Are you sure you are on the correct branch?

I would also assert that "sample_with_ancilla takes" and "seconds" are in the debug log and that there is a positive entry in the timing_stats.  Potentially, you could mock time.perf_counter in order to get an exact value for the time, but you probably don't need to do that for this example.

recirq/quantum_chess/quantum_board.py
```diff
@@ -133,6 +143,12 @@ def undo_last_move(self) -> bool:
                 return False
         return True
 
+    def record_time(self, t0, t1: Optional[float] = None) -> None:
+        if t1 is None:
+            t1 = time.perf_counter()
+        self.debug_log += (f"sample_with_ancilla takes {t1 - t0:0.4f} seconds.\n")
```
dstrain115, line 149: I wonder if we should make "sample_with_ancilla" as a parameter. Do you think we will later want to record other calls?

recirq/quantum_chess/quantum_board.py
```diff
@@ -102,6 +104,14 @@ def print_debug_log(self, clear_log: bool = True) -> None:
         print(self.debug_log)
         if clear_log:
             self.clear_debug_log()
+            
+    def clear_timing_stats(self) -> None:
```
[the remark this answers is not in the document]
madcpf: Thanks Doug. It's added.

recirq/quantum_chess/quantum_board.py
madcpf: Thanks Doug. The test is updated. I did not know about "pytest". Thanks for letting me know how to test the test file.

recirq/quantum_chess/quantum_board.py
madcpf: Thanks Doug. Now it becomes a parameter "action".

recirq/quantum_chess/quantum_board.py
```diff
@@ -255,13 +279,15 @@ def sample_with_ancilla(self, num_samples: int
...
                         f'{post_count} from post-selection\n')
+                    self.record_time("sample_with_ancilla", t0)
                     return (rtn, ancilla)
         else:
             rtn = [self.state] * num_samples
             self.debug_log += (
                 f'Discarded {error_count} from error mitigation '
                 f'{noise_count} from noise and {post_count} from post-selection\n'
             )
+        self.record_time("sample_with_ancilla", t0)
```
dstrain115, line 290: Let's use single quotes for consistency.

recirq/quantum_chess/quantum_board.py
```diff
@@ -133,6 +147,15 @@ def undo_last_move(self) -> bool:
...
         return True
 
+    def record_time(self, action: str, t0: float, t1: Optional[float] = None) -> None:
+        """Writes time span from t0 to t1 (if specified, otherwise the current time is used)
+        into the debug log and timing stats.
+        """
+        if t1 is None:
+            t1 = time.perf_counter()
+        self.debug_log += (f"{action} takes {t1 - t0:0.4f} seconds.\n")
+        self.timing_stats.append(t1 - t0)
```
dstrain115, line 157: Should we append a tuple of action/time instead?  (or alternatively, make timing_stats a dictionary of action to list of times?)

dstrain115 (commented):
Looking better!  Had one more suggestion.

recirq/quantum_chess/quantum_board.py
madcpf: Thanks! It's changed.

recirq/quantum_chess/quantum_board.py
madcpf: Thanks Doug! Yes that makes much sense. timing_stats is changed to dictionary, and the test has changed accordingly.

dstrain115 (approved):
LGTM.  Thanks for responding to all the comments!

PR #139, opened 2021-02-02
Fix batch_move to use correct gateset

dstrain115:
- QuantumProcessor uses a string to identify gateset instead
of a cirq object, so change the function to use this string.


mpharrigan:
lgtm; note we can probably start moving to `cirq.google.get_engine_sampler` and `cirg.google.get_engine_device` since those exist now

lingxz (approved):
Thanks!

Issue #133, opened 2021-01-19
Quantum Chess: Add ancillary qubits when mapping to hardware


dstrain115:
(More difficult issue)

When we try to map to hardware (especially for merge moves), sometimes the moves do not fit the topology of the device.  For instance, consider this move sequence:
 - Rook splits from a1 to b1
 - Rook on a2 moves to a3.
 - Rooks on a3 and b1 merge to b3.
 
This creates this tree: 
a1  - a2  -  a3
 |                |
b1      -      b3

This "pentagon" layout will not fit on the grid.  However, if we added a bonus ancillary qubit with SWAP gates at b2, it could fit in a 2x3 rectangle on the grid.  We need to be able to detect these problems and correct them by adding extra qubits on one of the paths.

Issue #132, opened 2021-01-19
Quantum Chess: use calibration metrics for better qubit picking


dstrain115:
We should explore using calibration metrics (https://quantumai.google/cirq/google/calibration) as a way of picking better qubits for quantum chess (i.e. picking qubits with lower sq/sqrt-iswap/readout errors).

See also #129 and #127 for similar issues.

Issue #127, opened 2021-01-19
Quantum Chess: Improve Qubit picking algorithm


dstrain115:
The qubit picking algorithm for quantum chess greedily picks the most connected qubit.  This is good if there is only one circuit transformation to do.  However, if there are two distinct "chains" of moves, this greedy algorithm may no longer work.  For instance, suppose the b1 knight splits on one side of the board, and the g1 knight splits on the other side of the board (and the two sides never interact).  It may be possible to put b1 circuit on the right side of the chip, and g1 circuit on the left side of the chip.  However, since the greedy algorithm put the b1 circuit right in the middle of the circuit, they can't both fit.

Another possibility that may be easier algorithmically (but possibly involve changing more code), is to split discrete chains of moves into different circuits, rather than try to fit both chains into the first circuit.  This could also slow down performance. 

The qubit picking algorithm for quantum chess is found here:
https://github.com/quantumlib/ReCirq/blob/master/recirq/quantum_chess/circuit_transformer.py#L58

PR #142, opened 2021-02-08
refactoring: make CircuitTransformer into an abstract interface

weinstein:
this way as additional transformations are implemented (see #127 #132 #133) we can implement additional transformations and swap them out more easily (ex for evaluation).

### DIFF
--- a/recirq/qaoa/placement.py
+++ b/recirq/qaoa/placement.py
@@ -6,7 +6,7 @@ from typing import Callable, Dict, List, Optional, Tuple
 import networkx as nx
 import numpy as np
 import pytket
-import pytket.cirq
+import pytket.extensions.cirq
 from pytket.circuit import Node, Qubit
 from pytket.passes import SequencePass, RoutingPass, PlacementPass
 from pytket.predicates import CompilationUnit, ConnectivityPredicate
@@ -87,7 +87,7 @@ def place_on_device(circuit: cirq.Circuit,
         initial_map: Initial placement of qubits
         final_map: The final placement of qubits after action of the circuit
     """
-    tk_circuit = pytket.cirq.cirq_to_tk(circuit)
+    tk_circuit = pytket.extensions.cirq.cirq_to_tk(circuit)
     tk_device = _device_to_tket_device(device)
 
     unit = CompilationUnit(tk_circuit, [ConnectivityPredicate(tk_device)])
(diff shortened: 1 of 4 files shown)


dstrain115 (approved):

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -22,19 +22,31 @@
 
 
 class CircuitTransformer:
+    """Abstract interface for circuit transformations.
+
+    For example: named -> grid qubit transformations.
```
wingers, line 27: nit: NamedQubit -> GridQubit transformations

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -22,19 +22,31 @@
...
+    """
+    def __init__(self):
```
wingers, line 29: nit: should we add `device` as a parameter?

wingers (approved):

recirq/quantum_chess/circuit_transformer.py
weinstein: IMO there's not much benefit to have the base class take the device and do nothing else with it besides storing a data member, even if all the implementations will end up doing this. If the implementations end up sharing some common functionality, there will probably be a cleaner way to share that.

wingers: sgtm 👍

recirq/qaoa/placement.py
```diff
@@ -103,7 +103,7 @@ def place_on_device(circuit: cirq.Circuit,
                      for n1, n2 in unit.initial_map.items()}
     final_map = {tk_to_cirq_qubit(n1): tk_to_cirq_qubit(n2)
              for n1, n2 in unit.final_map.items()}
-    routed_circuit = pytket.cirq.tk_to_cirq(unit.circuit)
+    routed_circuit = pytket.extensions.cirq.tk_to_cirq(unit.circuit)
```
dstrain115, line 106: Oh, I saw this on another PR, Nice job tracking this down!

dstrain115 (approved):

recirq/qaoa/placement.py
mpharrigan: ah in the future can you ping me or open this as a separate fix PR. This issue also exists in the notebooks and I had to re-discover the solution xref #148

weinstein: my bad! Will do next time.

mpharrigan: no worries! I was trying to complain at @dstrain115  :)

PR #144, opened 2021-02-11
Improve initial qubit mapping algorithm

wingers:
Fixes https://github.com/quantumlib/ReCirq/issues/127 and relates to https://github.com/quantumlib/ReCirq/issues/133.


recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,227 @@
+# Copyright 2020 Google
```
wingers, line 1: these tests were written as sanity checks, and i'm thinking they might be useful to keep around for debugging until after the swap update algorithm is implemented and proper tests for circuit transformations can be written?

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +307,391 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        """Returns a qubit that is a graph center.
+
+        Uses the Floyd-Warshall algorithm to calculate the length of the
+        shortest path between each pair of nodes. Then, finds the graph center
+        such that the length of the shortest path to the farthest node is the
+        smallest. Returns the first graph center if there are multiple.
+
+        Args:
+          g: A physical qubits graph or a logical qubits graph.
+        """
```
weinstein, line 440: should we move the floyd-warshall implementation into a separate file? That may be better organizationally. And since it is a fairly generic algorithm, it seems like something that we could add test coverage for independently of the rest of the initial mapping implementation

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +307,391 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        of the device.
+
+        Args:
+          circuit: The circuit to transform.
+        """
+        pg = self.build_physical_qubits_graph()
+        lg = self.build_logical_qubits_graph(circuit)
+        initial_mapping = self.calculate_initial_mapping(pg, lg)
+        # TODO: return the transformed circuit here instead.
+        return initial_mapping
```
weinstein, line 693: I think this doesn't type -- the initial mapping is a dict, but the return type is Circuit. We could transform_qubits() on the input circuit to get a circuit representing the initial mapping.

And if we did that, we could run the circuit_transformer_test cases on the DynamicLookAheadHeuristicCircuitTransformer as well as the existing one to see what happens -- I have a feeling that the initial mapping will be able to handle the small circuits in those test cases

wingers: the initial mapping algorithm ignores 1-qubit gates, and for some reason i had thought the swap update algorithm would take care of including them. but yah, i just updated the logic to handle 1-qubit gates (we're considering each as a single connected component).

i added tests for DynamicLookAheadHeuristicCircuitTransformer to the first three cases in circuit_transformer_test.py; since the heuristic used in this mapping algorithm is not exhaustive like in ConnectivityHeuristicCircuitTransformer, the last three cases will be flaky until we implement the swap update algorithm.

recirq/quantum_chess/circuit_transformer.py
wingers: i agree the floyd-warshall algorithm and the various bfs algorithms are fairly generic, but their implementation all have to be aware of whether it's graph param is a logical qubit graph (where edges contain moment info) or physical qubit graph or either, so i think it might make more sense to keep them in the same class.

but yah, it's def worth adding test coverage independently for these more generic algorithms; maybe we can keep the test files separate? kind of like what we have right now in dynamic_look_ahead_heuristic_circuit_transformer_test.py, but with more / better test cases?

recirq/quantum_chess/circuit_transformer.py
weinstein: I see. That is a bit unfortunate that even those small examples would require additional swaps, but good that we can add some extra test coverage!

recirq/quantum_chess/circuit_transformer.py
weinstein: Agreed, lets add smaller focused files if we can. Besides the tests I'm also a bit concerned about the potential for circuit_transformer.py to explode in size if we add more transformers, but should be fine for now.

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,227 @@
...
+logical2 = {
+    a0: [(a2, 0), (a1, 1), (a3, 6)],
+    a1: [(a0, 1), (a2, 3), (a7, math.inf)],
+    a2: [(a0, 0), (a1, 3), (a3, 5)],
+    a3: [(a2, 5), (a0, 6)],
+    a4: [(a5, 0), (a6, math.inf)],
+    a5: [(a4, 0)],
+    a6: [(a7, 0), (a4, math.inf)],
+    a7: [(a6, 0), (a1, math.inf)],
+}
```
weinstein, line 47: what is "logical" and "logical2"? Expected outputs for some particular test case(s)?
I think it may be better to inline those expectations in the test body so that (1) its clearer to readers what the meaning of these dicts is and (2) readers of the test body won't have to refer back-and-forth to these definitions at a different place in the file

ditto with the 'physical' dict below which I think is used in a similar way for some tests input

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,227 @@
...
+one_one = cirq.GridQubit(1, 1)
+one_two = cirq.GridQubit(1, 2)
+one_three = cirq.GridQubit(1, 3)
+one_four = cirq.GridQubit(1, 4)
+one_five = cirq.GridQubit(1, 5)
+one_six = cirq.GridQubit(1, 6)
+one_seven = cirq.GridQubit(1, 7)
+one_eight = cirq.GridQubit(1, 8)
+one_nine = cirq.GridQubit(1, 9)
+one_ten = cirq.GridQubit(1, 10)
```
weinstein, line 70: would it be cleaner to store these in an array/dict and construct them in a loop?
ex something like:
```
grid_qubits = dict(
    ('{row}_{col}', cirq.GridQubit(row, col))
    for row in range(2) for col in range(11))
    
... grid_qubits['1_10'] ...
```

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +307,391 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        g: Dict[cirq.Qid, List[Tuple[cirq.Qid, int]]],
+        component: Deque[cirq.Qid],
+    ) -> cirq.Qid:
+        """Returns the least connected qubit.
+
+        Args:
+          g: A logical qubits graph.
+          component: A deque of qubits belonging to the same component.
+        """
+        qubit = None
```
weinstein, line 354: if you want, you can use min() with an ordering function to compute this directly:
```
return min(component, key=lambda q: len(g[q]))
```

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: done, and added more test cases!

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: done!

recirq/quantum_chess/circuit_transformer.py
wingers: nice, updated!

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+
+        return g
+
+    def find_graph_center(
+        self,
+        g: Union[
+            Dict[cirq.GridQubit, List[cirq.GridQubit]],
+            Dict[cirq.Qid, List[Tuple[cirq.Qid, int]]],
+        ],
+    ) -> Union[cirq.GridQubit, cirq.Qid]:
```
weinstein, line 437: nit: I think GridQubit inherits from Qid, so Union[GridQubit, Qid] is just Qid

recirq/quantum_chess/circuit_transformer.py
weinstein: by the way, I realized another reason it would be nice to refactor out the floyd-warshall implementation: the swap update step needs the precomputed node-to-node shortest path distances, too. We can leave that for a future pull request though.

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        added to the result.
+
+        Args:
+          mapping: The current mapping of logical qubits to physical qubits.
+          lg: A logical qubits graph.
+          lq: The logical qubit from which to find reference qubits.
+        """
+        qubits = []
+        neighbors_sorted_by_moment = sorted(lg[lq], key=lambda x: x[1])
+        for neighbor in neighbors_sorted_by_moment:
```
weinstein, line 542: nit, this can be simplified:
```
for neighbor in neighbors_sorted_by_moment.keys():
```

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        pq: cirq.GridQubit,
+    ) -> List[cirq.GridQubit]:
+        """Returns a list of physical qubits available to be mapped.
+
+        Uses level order traversal until a level with free adjacent node(s) is
+        found.
+
+        Args:
+          mapped: The set of currently mapped physical qubits.
+          lg: A physical qubits graph.
```
weinstein, line 563: rename `lg` -> `pg` and `lq` -> `pq` to match the arg names?

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        traversal_order = self.traverse(lg, lg_center)
+
+        while traversal_order:
+            lq = traversal_order.popleft()
+            if lq == lg_center:
+                continue
+            pq = None
+            reference_qubits = self.find_reference_qubits(mapping, lg, lq)
+            ref_q = reference_qubits[0]
+            candidate_qubits = self.find_candidate_qubits(mapping.values(), pg, ref_q)
```
weinstein, line 647: personally I found it a bit confusing with ref_q being shadowed in the loop below -- could inline reference_qubits[0] as the find_candidate_qubits() reference qubit arg instead

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+                    distances = defaultdict(list)
+                    for cand_q in candidate_qubits:
+                        d = self.find_shortest_path_distance(pg, ref_q, cand_q)
+                        distances[d].append(cand_q)
+                    nearest_candidate_qubits = None
+                    min_dist = math.inf
+                    for dist in distances:
+                        if dist < min_dist:
+                            min_dist = dist
+                            nearest_candidate_qubits = distances[dist]
```
weinstein, line 662: nit: could we replace this block with a min() call? It looks like
```
min_dist = min(distances.keys())
candidate_qubits = distances[min_dist]
```

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -33,59 +33,64 @@
...
 def test_single_qubit_ops(device):
-    transformer = ct.ConnectivityHeuristicCircuitTransformer(device)
     c = cirq.Circuit(cirq.X(a1), cirq.X(a2), cirq.X(a3))
-    transformer.qubit_mapping(c)
-    c = transformer.transform(c)
-    device.validate_circuit(c)
+    t = ct.ConnectivityHeuristicCircuitTransformer(device)
+    device.validate_circuit(t.transform(c))
+    t = ct.DynamicLookAheadHeuristicCircuitTransformer(device)
+    device.validate_circuit(t.transform(c))
```
weinstein, line 40: I think we can make pytest parameterize the transformer type for these tests, similar to what we do for the device

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,330 @@
...
+def test_get_least_connected_qubit():
+    t = ct.DynamicLookAheadHeuristicCircuitTransformer(cirq.google.Foxtail)
+    g = {
+        0: [1],
+        1: [0, 2],
+        2: [1],
+        3: [4],
+        4: [3],
+    }
+    assert t.get_least_connected_qubit(g, deque([0, 1, 2])) == 0
```
weinstein, line 162: is this test overspecified -- should we assert this result is in {0, 2} instead? I'm not sure how get_least_connected_qubit should break ties (or if we should make this test care)

weinstein (approved):

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+
+        # Build an adjacency graph based on the circuit.
+        for i, m in enumerate(circuit):
+            moment_index = i
+            for op in m:
+                if len(op.qubits) == 1:
+                    q = op.qubits[0]
+                    if q not in g:
+                        g[q] = []
+                if len(op.qubits) == 2:
```
dstrain115, line 392: Should we change this to `elif len(op.qubits) == 2` and then add an else that handles len(op.qubits) > 2 (possibly by throwing a ValueError?  It might be a good idea to make sure we skip measurement gates properly too.

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        The transformed circuit satisfies all physical adjacency constraints
+        of the device.
+
+        Args:
+          circuit: The circuit to transform.
+        """
+        pg = self.build_physical_qubits_graph()
+        lg = self.build_logical_qubits_graph(circuit)
+        initial_mapping = self.calculate_initial_mapping(pg, lg)
+        # TODO: change this to use the result of the SWAP update algorithm.
```
dstrain115, line 695: Add owner or link to issue for TODO.

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -33,59 +33,64 @@
...
 
 
 @pytest.mark.parametrize('device',
                          (cirq.google.Sycamore23, cirq.google.Sycamore))
 def test_single_qubit_with_two_qubits(device):
```
dstrain115, line 45: What does single_qubit_with_two_qubits mean?  Can we rename this to be more descriptive?

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -33,59 +33,64 @@
...
 def test_disconnected(device):
-    transformer = ct.ConnectivityHeuristicCircuitTransformer(device)
     c = cirq.Circuit(qm.split_move(a1, a2, a3), qm.split_move(a3, a4, d1),
                      qm.split_move(b1, b2, b3), qm.split_move(c1, c2, c3))
-    transformer.qubit_mapping(c)
-    device.validate_circuit(transformer.transform(c))
+    t = ct.ConnectivityHeuristicCircuitTransformer(device)
+    device.validate_circuit(t.transform(c))
+#    t = ct.DynamicLookAheadHeuristicCircuitTransformer(device)
+#    device.validate_circuit(t.transform(c))
```
dstrain115, line 73: REmove commented code.  Same below.

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,330 @@
+# Copyright 2020 Google
```
dstrain115, line 1: nit: It's 2021 now.

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -33,59 +33,64 @@
...
-    transformer = ct.ConnectivityHeuristicCircuitTransformer(device)
     c = cirq.Circuit(qm.split_move(a1, a2, b1), qm.split_move(a2, a3, b3),
                      qm.split_move(b1, c1, c2), qm.normal_move(c1, d1),
                      qm.normal_move(a3, a4), qm.merge_move(a4, d1, a1))
-    transformer.qubit_mapping(c)
-    device.validate_circuit(transformer.transform(c))
+    t = ct.ConnectivityHeuristicCircuitTransformer(device)
+    device.validate_circuit(t.transform(c))
+#    t = ct.DynamicLookAheadHeuristicCircuitTransformer(device)
+#    device.validate_circuit(t.transform(c))
```
dstrain115, line 96: Should we add tests for some of the public functions?  For instance, adding functionality to test least_connected_qubit should be simple enough.

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
```diff
@@ -0,0 +1,330 @@
...
+import recirq.quantum_chess.circuit_transformer as ct
+
+a0 = cirq.NamedQubit('a0')
+a1 = cirq.NamedQubit('a1')
+a2 = cirq.NamedQubit('a2')
+a3 = cirq.NamedQubit('a3')
+a4 = cirq.NamedQubit('a4')
+a5 = cirq.NamedQubit('a5')
+a6 = cirq.NamedQubit('a6')
+a7 = cirq.NamedQubit('a7')
```
dstrain115, line 29: Optional:  I think the style guide would have these  e capitalized (i.e. A7), but this goes against chess algebraic style, so I am not sure which is better.

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
dstrain115: I think we should probably keep these tests, but maybe split out DynamicLookAheadHeuristicCircuitTransformer so that it matches the name of the test file.

recirq/quantum_chess/circuit_transformer.py
wingers: sg, i refactored it out to a separate function!

recirq/quantum_chess/circuit_transformer.py
wingers: updated!

recirq/quantum_chess/circuit_transformer.py
wingers: `neighbors_sorted_by_moment` is a list but updated to make this a bit cleaner.

recirq/quantum_chess/circuit_transformer.py
wingers: done!

recirq/quantum_chess/circuit_transformer.py
wingers: updated!

recirq/quantum_chess/circuit_transformer.py
wingers: done!

recirq/quantum_chess/circuit_transformer_test.py
wingers: done!

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: yah agreed the test shouldn't need to know the implementation, updated.

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
         return circuit.transform_qubits(lambda q: self.mapping[q])
 
 
+class DynamicLookAheadHeuristicCircuitTransformer(CircuitTransformer):
+    """Optimizer that transforms a circuit to satify a device's constraints.
+
+    This implements the initial mapping algorithm and the SWAP update algorithm
```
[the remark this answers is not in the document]
wingers: done!

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -305,6 +317,384 @@ def transform(self, circuit: cirq.Circuit) -> cirq.Circuit:
...
+        self,
+    ) -> Dict[cirq.GridQubit, List[cirq.GridQubit]]:
+        """Returns an adjacency graph of physical qubits of the device.
+
+        Each edge is bidirectional, and represents a valid two-qubit gate.
+        """
+        g = defaultdict(list)
+        for q in self.device.qubit_set():
+            neighbors = [n for n in q.neighbors() if n in self.device.qubit_set()]
+            for n in neighbors:
```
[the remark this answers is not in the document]
wingers: updated!

recirq/quantum_chess/circuit_transformer.py
wingers: sg, updated to throw a `ValueError` when `len(ops.qubits) > 2`; also added a check to skip measurement gates.

recirq/quantum_chess/circuit_transformer.py
wingers: done!

recirq/quantum_chess/circuit_transformer_test.py
wingers: renamed!

recirq/quantum_chess/circuit_transformer_test.py
wingers: done!

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: updated!

recirq/quantum_chess/circuit_transformer_test.py
wingers: yah, tests are added in `initial_mapping_utils_test.py` (which was previously `dynamic_look_ahead_heuristic_circuit_transformer_test.py`).

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: i'm leaning towards using lower case for consistency since that's currently used in https://github.com/quantumlib/ReCirq/blob/master/recirq/quantum_chess/constants.py and https://github.com/quantumlib/ReCirq/blob/master/recirq/quantum_chess/circuit_transformer_test.py.

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
wingers: ack, i refactored the initial mapping logic out to `initial_mapping_utils.py` and renamed this to `initial_mapping_utils_test.py`

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -30,62 +30,67 @@
...
+def test_three_split_moves(transformer, device):
     c = cirq.Circuit(qm.split_move(a1, a2, b1), qm.split_move(a2, a3, b3),
                      qm.split_move(b1, c1, c2))
-    transformer.qubit_mapping(c)
-    device.validate_circuit(transformer.transform(c))
+    t = transformer(device)
+    device.validate_circuit(t.transform(c))
 
 
+@pytest.mark.parametrize('transformer', [ct.ConnectivityHeuristicCircuitTransformer])
```
dstrain115, line 68: Why are we parameterizing if it only has one option?  Same with several functions below.

recirq/quantum_chess/initial_mapping_utils_test.py
```diff
@@ -0,0 +1,386 @@
...
+def test_get_least_connected_qubit():
+    g = {
+        0: [1],
+        1: [0, 2],
+        2: [1],
+        3: [4],
+        4: [3],
+    }
+    assert imu.get_least_connected_qubit(g, deque([0, 1, 2])) in {0, 2}
+    assert imu.get_least_connected_qubit(g, deque([3, 4])) in {3, 4}
```
dstrain115, line 161: Do we want to have a test case that is less ambiguous?

dstrain115 (approved):

recirq/quantum_chess/initial_mapping_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    m: Dict[cirq.Qid, int],
+) -> List[List[int]]:
+    """ Returns a matrix of the shortest distance between each pair of nodes.
+
+    Implements the Floyd–Warshall algorithm.
+
+    Args:
+      g: A physical qubits graph or a logical qubits graph.
+      v: The size of the graph.
+      m: A mapping of qubit to index.
```
weinstein, line 136: would the interface be simpler if we keyed things by Qid directly and returned a dict from Pair[Qid, Qid] -> int instead of requiring an intermediate Qid -> index mapping? Then the only args we'd need to provide would be the graph dict

recirq/quantum_chess/dynamic_look_ahead_heuristic_circuit_transformer_test.py
cantwellc: I'd agree on lower, for consistency with chess.

cantwellc (approved):

recirq/quantum_chess/circuit_transformer_test.py
wingers: added `DynamicLookAheadHeuristicCircuitTransformer` to the list now that it's updated to use `SwapUpdater`.

recirq/quantum_chess/initial_mapping_utils_test.py
wingers: added!

recirq/quantum_chess/initial_mapping_utils.py
wingers: updated and it's much cleaner now!

weinstein: just need to update the return type hint to something like Dict[Tuple[Qid, Qid], int]  (who/what is using these type hints anyways...?)

weinstein (approved):

recirq/quantum_chess/initial_mapping_utils.py
wingers: ty for catching that! i'm guessing they're for eventual docs generation...?

PR #145, opened 2021-02-13
Fix the queen's position in the 'double occupancy' section.

PawelPamula:
If we measured the king to be at 'b2', the queen would not be allowed to move and thus would stay in the initial position - 'a3'.


dstrain115 (approved):
Good catch!  Thanks for fixing it!

PR #146, opened 2021-02-16
implement swap update algorithm for adjacency constraint resolution when placing circuits on hardware

weinstein:
This is the second half of the algorithm described in [this paper](https://ieeexplore.ieee.org/abstract/document/8976109) and is related to #127 and #133 

The swap updater takes an initial logical -> device qubit mapping and adds additional qubits and swap operations to ensure all gates operate on adjacent qubits. When combined with the initial placement implementation from #144 we'll get a complete circuit placement algorithm.


recirq/quantum_chess/swap_update_transformer.py
```diff
@@ -0,0 +1,180 @@
...
+        If the updater already completed, does nothing.
+
+        Returns:
+          the generated operations on physical GridQubits in the final circuit
+        """
+        while not self.dlists.all_empty():
+            yield from self.update_iteration()
+
+
+class SwapUpdateTransformer(ct.CircuitTransformer):
```
wingers, line 155: wondering what your thoughts are on how we will combine our changes?

the initial mapping logic i implemented are technically all util functions and can't be used as a standalone transformer since it assumes some swap update algorithm will update its mapping after.

i understand `SwapUpdater` can be used with any arbitrary initial mapping, but perhaps we should treat it as a util class and other transformer implementations can explicitly call it if they choose to? please correct me if i'm wrong, but my understanding of the `CircuitTransformer` abstraction is that it'll take a device and a circuit and automagically return a circuit that will work on the device without the caller having to know to sometimes create an initial mapping first.

i'll move all the logic in `DynamicLookAheadHeuristicCircuitTransformer` (aside from `transform`) to a separate util file in my PR. we could then either rename this class to `DynamicLookAheadHeuristicCircuitTransformer` or just leave it as is, and update the `transform` method to call the (to be refactored) initial mapping utils and `SwapUpdater`? or we could then also move this class to `circuit_transformer.py` since it'll be relatively tiny and have this file contain `SwapUpdater` related logic for other transformers to use?

we can ofc do the refactoring in a subsequent PR but just wanted to make sure we are on the same page!

wingers (commented):
this is great! so exciting to see it all coming together :)

recirq/quantum_chess/swap_update_transformer.py
weinstein: Actually, I was planning to use the SwapUpdater directly in your DynamicLookAheadHeuristicCircuitTransformer implementation; it sounds like we're on the same page there.

Initially I had added this stand-alone transformer because CircuitTransformer sounds like a more generic circuit -> circuit mapper without any other constraints on the inputs or outputs, and who knows, maybe we'll end up with other kinds of initial placements that would benefit from applying the swap updates.
But I changed my mind -- QuantumBoard requires CircuitTransformer output to be valid and runnable, even if that's not explicit in the interface. And even if we do end up with other kinds of initial placements, there's no reason we can't use SwapUpdater directly for those either. Having the transformer doesn't really add much.

recirq/quantum_chess/swap_update_transformer.py
```diff
@@ -0,0 +1,180 @@
...
+          circuit: the input circuit of GridQubits to modify
+        Returns:
+          the modified circuit with ISWAPs inserted and operations mapped to
+          physical GridQubits.
+        """
+        for q in circuit.all_qubits():
+            assert q in self.initial_mapping
+        updater = SwapUpdater(circuit, circuit.device.qubit_set(),
+                              self.initial_mapping)
+        return cirq.Circuit(updater.add_swaps(), device=circuit.device)
```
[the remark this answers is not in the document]
weinstein: Sounds reasonable (wonder if there are any plans to deprecate that param / update Circuit in future cirq releases?)

recirq/quantum_chess/swap_update_transformer.py
```diff
@@ -0,0 +1,180 @@
...
+        least one of the given physical gates.
+
+        Args:
+          gates: the list of gates to consider which operate on GridQubits
+        """
+        for gate in gates:
+            for gate_q in gate.qubits:
+                yield from (
+                    (gate_q, swap_q)
+                    for swap_q in self._adjacent_qubits(gate_q)
```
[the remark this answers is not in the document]
weinstein: Ah, perfect! Thank you

recirq/quantum_chess/swap_update_transformer.py
wingers: awesome sg!

wingers (approved):

recirq/quantum_chess/swap_updater.py
```diff
@@ -0,0 +1,140 @@
...
+    return q1.is_adjacent(q2)
+
+
+def generate_decomposed_swap(
+        q1: cirq.Qid, q2: cirq.Qid) -> Generator[cirq.Operation, None, None]:
+    """Generates a SWAP operation using sqrt-iswap gates."""
+    yield cirq.ISWAP(q1, q2)**0.5
+    yield cirq.ISWAP(q1, q2)**0.5
+    yield cirq.ISWAP(q1, q2)**0.5
+    yield cirq.ISWAP(q1, q2)**0.5
```
dstrain115, line 46: As evidenced by the above comment, we should test this function.

recirq/quantum_chess/swap_updater.py
```diff
@@ -0,0 +1,140 @@
...
+        # If all the active gates in this pass were added to the final circuit,
+        # then we have nothing left to do until we make another pass and get the
+        # newly active gates.
+        if not active_physical_gates:
+            return
+
+        candidates = set(self.generate_candidate_swaps(active_physical_gates))
+        # This could happen on irregular topologies or when some device
+        # qubits are disallowed.
+        assert candidates, 'no swaps found that will improve the circuit'
```
dstrain115, line 123: Will this also happen if the mapping is already optimal?  I am not sure whether asserting here is the correct thing to do.

recirq/quantum_chess/swap_updater_test.py
```diff
@@ -0,0 +1,117 @@
...
+q = list(cirq.NamedQubit(f'q{i}') for i in range(6))
+
+# Qubits corresponding to figure 9 in
+# https://ieeexplore.ieee.org/abstract/document/8976109.
+# Coupling graph looks something like:
+#   Q0 = Q1 = Q2
+#   ||   ||   ||
+#   Q3 = Q4 = Q5
+FIGURE_9A_PHYSICAL_QUBITS = list(
+    cirq.GridQubit(row, col) for row in range(2) for col in range(3))
```
dstrain115, line 18: You can also use cirq.GridQubit.rect()

recirq/quantum_chess/swap_updater_test.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def test_decomposed_swaps():
+    Q = FIGURE_9A_PHYSICAL_QUBITS
+    initial_mapping = dict(zip(q, Q))
+    updater = SwapUpdater(FIGURE_9A_CIRCUIT, Q, initial_mapping,
+                          generate_decomposed_swap)
+    # First iteration adds a swap between Q0 and Q1.
+    # generate_decomposed_swap implements that swap operation as four
+    # sqrt-iswaps.
```
dstrain115, line 111: Instead of testing the implementation, we should test the effect.  For instance, that the effect is equivalent to swapping Q[0] and Q[1] (via cirq.unitary or some other method).

recirq/quantum_chess/swap_updater.py
weinstein: if the mapping were already optimal, we'd always fall out from the `if not active_physical_gates` condition just before this (I will add some test coverage for already-optimal circuits regardless).

recirq/quantum_chess/swap_updater_test.py
weinstein: great idea.

Related question: how is the unitary matrix index determined for a circuit's qubits? For some of these tests, it would be cool to compare the logical circuit's unitary before and after adding swaps; they should be the same, but only after ignoring any ancillary qubits.

recirq/quantum_chess/mcpe_utils.py
```diff
@@ -0,0 +1,194 @@
...
+    """Data structure representing a 1:1 map between logical and physical GridQubits.
+
+    Args:
+      initial_mapping: initial logical-to-physical qubit map.
+    """
+    def __init__(self, initial_mapping: Dict[cirq.Qid, cirq.GridQubit] = {}):
+        self.logical_to_physical = initial_mapping
+        self.physical_to_logical = {v: k for k, v in initial_mapping.items()}
+
+    def insert(self, logical_q: cirq.Qid, physical_q: cirq.GridQubit) -> None:
```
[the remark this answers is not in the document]
weinstein: Removed it.

recirq/quantum_chess/mcpe_utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Args:
+          swap_q1: the source qubit to swap
+          swap_q2: the target qubit to swap
+          gates: the dependency list of gate operations on logical qubits
+          mapping: the mapping between logical and physical qubits for gates
+        """
+        total_cost = 0
+        for gate in gates:
+            assert len(gate.qubits) <= 2
```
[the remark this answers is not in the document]
weinstein: I went ahead and changed all the `assert`s in mcpe_utils and swap_updater to raise ValueError instead

recirq/quantum_chess/mcpe_utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+import cirq
+
+
+def manhattan_dist(q1: cirq.GridQubit, q2: cirq.GridQubit) -> int:
+    """Returns the Manhattan distance between two GridQubits.
+
+    On grid devices this is the shortest path length between the two qubits.
+    """
+    return abs(q1.row - q2.row) + abs(q1.col - q2.col)
```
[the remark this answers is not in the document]
weinstein: Yes. Added some more tests to `mcpe_utils_test.py` for these

recirq/quantum_chess/mcpe_utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Note that this returns >0 if the distance would decrease and <0 if it would
+    increase, which is somewhat counter-intuitive.
+
+    Args:
+      swap_qubits: the pair of qubits to swap
+      gate_qubits: the pair of qubits that the gate operates on
+    """
+    gate_after = map(swap_map_fn(*swap_qubits), gate_qubits)
+    # TODO: Using manhattan distance only works for grid devices when all qubits
```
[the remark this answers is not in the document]
weinstein: Sounds good, happy to open issues for stuff like this. I opened #149

recirq/quantum_chess/swap_updater.py
```diff
@@ -0,0 +1,140 @@
...
+    assert len(gate.qubits) <= 2
+    if len(gate.qubits) < 2:
+        return True
+    q1, q2 = gate.qubits
+    return q1.is_adjacent(q2)
+
+
+def generate_decomposed_swap(
+        q1: cirq.Qid, q2: cirq.Qid) -> Generator[cirq.Operation, None, None]:
+    """Generates a SWAP operation using sqrt-iswap gates."""
```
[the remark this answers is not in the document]
weinstein: Ouch, good catch. I switched it to use optimize_for_sycamore() for simplicity but LMK if you feel strongly about using the conversion function directly. Or maybe it would be better/simpler to take an optional PointOptimizer instead of a swap_factory callable?

recirq/quantum_chess/swap_updater_test.py
weinstein: perfect, thanks.

mpharrigan:
Is this of general utility? Should this be upstreamed to cirq?

rmlarose:
> Is this of general utility? Should this be upstreamed to cirq?

FWIW, I had a university partner ask about routing code. I pointed them to cirq.contrib and here, and found it odd to point someone to quantum chess for this.

dstrain115:
> FWIW, I had a university partner ask about routing code. I pointed them to cirq.contrib and here, and found it odd to point someone to quantum chess for this.

Maybe we could try it out in Quantum chess and export it to cirq after some testing?

weinstein:
> > FWIW, I had a university partner ask about routing code. I pointed them to cirq.contrib and here, and found it odd to point someone to quantum chess for this.
> 
> Maybe we could try it out in Quantum chess and export it to cirq after some testing?

That would be my preference as well. Happy to file a separate PR against cirq.contrib later but I'd prefer to do some evaluation first and iron out some of the kinks.

rmlarose:
> Maybe we could try it out in Quantum chess and export it to cirq after some testing?

Seems reasonable to me!

recirq/quantum_chess/swap_updater.py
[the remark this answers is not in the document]
dstrain115: Calling optimized_for_sycamore might be less efficient, but it's probably fine for now.

dstrain115 (approved):

cantwellc (approved):
lgtm.

PR #148, opened 2021-02-17
Fix PyTket

mpharrigan:
There's something fishy going on with the pytket notebook. When running in the doc build system it says

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-2f5bab2028d6> in <module>
     13     return pytket.device.Device({}, {}, arc)
     14 
---> 15 tk_circuit = pytket.cirq.cirq_to_tk(circuit)
     16 tk_device = _device_to_tket_device()

AttributeError: module 'pytket' has no attribute 'cirq'
AttributeError: module 'pytket' has no attribute 'cirq'
```

and when I try to execute it in colab it complains about a bad sympy version

```
ERROR: pytket 0.7.1.1 has requirement sympy~=1.6, but you'll have sympy 1.1.1 which is incompatible.
```

and when I run the import pytket part, the kernel **dies** with some information in the logs:

```
AttributeError: module 'sympy.core.core' has no attribute 'numbers'
```

### DIFF
--- a/docs/qaoa/routing_with_tket.ipynb
+++ b/docs/qaoa/routing_with_tket.ipynb
@@ -93,7 +93,7 @@
     "try:\n",
     "    import recirq\n",
     "except ImportError:\n",
-    "    !pip install -q git+https://github.com/quantumlib/ReCirq"
+    "    !pip install -q git+https://github.com/quantumlib/ReCirq sympy~=1.6"
    ]
   },
   {
--- a/docs/qaoa/routing_with_tket.ipynb
+++ b/docs/qaoa/routing_with_tket.ipynb
@@ -93,7 +93,7 @@
     "try:\n",
     "    import recirq\n",
     "except ImportError:\n",
-    "    !pip install git+https://github.com/quantumlib/ReCirq"
+    "    !pip install -q git+https://github.com/quantumlib/ReCirq"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


rmlarose:
Still failing on colab for me w/ same error:

`ERROR: pytket 0.7.1.1 has requirement sympy~=1.6, but you'll have sympy 1.1.1 which is incompatible.`

Can this be fixed? 

https://colab.research.google.com/github/mpharrigan/recirq/blob/2021-02-pytket/docs/qaoa/routing_with_tket.ipynb#scrollTo=8cBvm5NrNCXt

mpharrigan:
I know it says "ERROR" but does it actually prevent the notebook from working?

On colab, if I put `pip install pytket~=0.7 pytket-cirq~=0.7 sympy~=1.6` it worked fine. I truly don't understand why the pip resolver doesn't upgrade sympy. Even in the error message it says "you'll have". I think it doesn't want to upgrade the system sympy installed in colab?

rmlarose:
> I know it says "ERROR" but does it actually prevent the notebook from working?

It ran until the cell importing pytket, at which point "Your session crashed for an unknown reason."

rmlarose:
Latest commit pinning `sympy~=1.6` runs through on colab.

rmlarose:
> I truly don't understand why the pip resolver doesn't upgrade sympy. Even in the error message it says "you'll have". I think it doesn't want to upgrade the system sympy installed in colab?

I think so, I'm also mesmerized by it though. I tried multiple combinations of `pip install -q git+https://github.com/quantumlib/ReCirq` with `--ignore-installed`, `--upgrade`, and `--force-reinstall` flags. I managed to upgrade sympy, but also a bunch of other packages got upgraded and caused more issues...

mpharrigan:
Good to know. We'll go with this current fix for now!

PR #150, opened 2021-02-23
Store repetition number when caching accumulations

### LINKED ISSUE #112: Quantum Chess: caching accumulations ignores number of samples
QuantumBoard.get_probability_distributions caches the probability accumulations:

```
        if not self.accumulations_valid:
            self._generate_accumulations(repetitions)
```

However, this is incorrect if the number of samples changes.  For instance, calling get_probability_distributions(repetitions=1) will generate probabilities of 1.0 or 0.0.  If I then call get_probability_distributions(repetitions=1000), it will return this same distribution again, instead of the more fine-grained distribution.

lingxz:
Fixes #112


dstrain115 (approved):

dstrain115:
LGTM.

PR #151, opened 2021-03-02
make SwapUpdater precompute all the qubit-to-qubit shortest path lengths

### LINKED ISSUE #149: Update SwapUpdater to use computed shortest path length instead of Manhattan distance
The current SwapUpdater implementation (#146) takes a short cut and uses manhattan distance when computing the shortest GridQubit-to-GridQubit distances.

But the manhattan distance is only the same as the shortest path length on the device connectivity graph when there are no unusable qubits (no holes, hanging strips, or disconnected qubits).

We should update the SwapUpdater to precompute shortest path lengths on the device connectivity graph (ex using the Floyd-Warshall algorithm, which we're already using for the initial circuit placement) and use those distances instead of manhattan distance. Otherwise, circuit placement will break in the presence of pruned qubits (see also #129).

weinstein:
this makes the swap updater work on devices that aren't regular grids and fixes #149


recirq/quantum_chess/swap_updater.py
```diff
@@ -39,6 +39,37 @@ def _satisfies_adjacency(gate: cirq.Operation) -> bool:
...
+        between them (disconnected qubits will be absent).
+    """
+    # Do BFS starting from each qubit and collect the shortest path lengths as
+    # we go.
+    # For device graphs where all edges are the same (where each edge can be
+    # treated as unit length) repeated BFS is O(v**2) and is faster than
+    # the Floyd-Warshall algorithm which is O(v**3).
+    # This won't work if in the future we figure out a way to incorporate edge
+    # weights (for example in order to give negative preference to gates with poor
+    # calibration metrics).
```
dstrain115, line 58: Optional: should we put this comment into the docstring where it is more visibile?
(Or do you think it is so implementation-specific that it belongs in the code?)

recirq/quantum_chess/swap_updater.py
```diff
@@ -39,6 +39,37 @@ def _satisfies_adjacency(gate: cirq.Operation) -> bool:
...
+    """Precomputes the shortest path length between each pair of qubits.
+
+    Returns:
+        dictionary mapping a pair of qubits to the length of the shortest path
+        between them (disconnected qubits will be absent).
+    """
+    # Do BFS starting from each qubit and collect the shortest path lengths as
+    # we go.
+    # For device graphs where all edges are the same (where each edge can be
+    # treated as unit length) repeated BFS is O(v**2) and is faster than
```
dstrain115, line 54: Since this is exponential in the number of edges, should we consider aborting if the number of edges is above a sanity limit?  If you believe the hype, we are not _that_ far away from much larger qubit counts.

dstrain115 (approved):
LGTM with minor comments.

recirq/quantum_chess/swap_updater.py
weinstein: I think it's worth adding a comment about the runtime complexity to the docstring, but I'd avoid adding more details about the implementation beyond that.

recirq/quantum_chess/swap_updater.py
weinstein: each round of bfs visits each vertex and edge at most once so O(V+E), and we do bfs V times. That's O(V**2 + V E) ie quadratic in the number of edges and vertices, not exponential

Assuming with bigger devices qubits are still likely to be locally connected (like grids where the device graph is sparse and edges per vertex bounded by a constant) then E is O(V) and this function is O(V**2) overall. Not sure if that assumption will withstand the test of time, but at least on a graph of 'GridQubit' it seems like a safe assumption. Even a fully connected graph would have at most V**2 edges making repeated bfs cubic in the number of qubits O(V**3) not exponential.

We're precomputing O(V**2) vertex-to-vertex distance values so it is highly unlikely that we'll be able to do any better on sparse device graphs or grids :)

PR #155, opened 2021-03-04
Update comments in quantum_moves.py

madcpf:


recirq/quantum_chess/quantum_board_test.py
```diff
@@ -67,6 +67,7 @@ def test_initial_state(board):
         assert probs[bit] == 0.0
 
 
+
```
dstrain115, line 70: I think the standards are two blank lines in between top-level functions.  We don't need a third line.

dstrain115 (commented):
Looks good so far.

PR #161, opened 2021-03-26
Fix DFS functions in circuit_transformer.py

madcpf:
- Fixes DFS in qubits_within() by adding a set (visited). Adds a test for this.
- Fixes DFS in edges_within() by adding a set (visited). Adds a test for this.
- Fixes DFS in map_helper() by adding a list (nodes_trying, using as a stack). After fixing the 2 DFSs above, the last test in circuit_transformer_test.py broke, which led me to this 3rd bug in DFS. Now that test passes.
In a normal DFS we don't need this extra variable. But what's special here is that

call this the 1st run:
```
        success = self.map_helper(node_to_map, mapping, available_qubits,
                                  graph, nodes_trying)

```
call this the 2nd run:
```
        if success:
            success = self.map_helper(cur_node, mapping, available_qubits,
                                      graph, nodes_trying)

```
When the 1st run returns success but the second run returns failure, all nodes involved in 1st and 2nd run are supposed to be reversed in mapping/available_qubits, while before this PR the behavior is to only reverse the top-level ones (i.e. node_to_map, node_to_try). The result is that the loop after this will have less available_qubits to try for, and may end up with a failure in creating the mapping. One perfect example is the last testing case in circuit_transformer_test.py.
- Updates comment for find_start_node().


dstrain115:
Adding a few others who have recently touched this function.

recirq/quantum_chess/bit_utils.py
```diff
@@ -45,6 +45,16 @@ def num_ones(n: int) -> int:
...
+def bit_ones(n: int) -> List[int]:
+    """Indices of ones in the binary representation of n."""
+    indices = []
+    index = 0
+    while n > 0:
+        if n % 2 == 1:
+            indices.append(index)
+        n >>= 1
+        index += 1
+    return indices
```
dstrain115, line 57: Wasn't this in your other PR?  Can you merge so that this is picked up?

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -49,7 +49,8 @@ def __init__(self, device: cirq.Device):
         self.qubit_list = device.qubits
 
     def qubits_within(self, depth: int, qubit: cirq.GridQubit,
-                      qubit_list: Iterable[cirq.GridQubit]) -> int:
+                      qubit_list: Iterable[cirq.GridQubit],
+                      visited: Set[cirq.GridQubit]) -> int:
```
dstrain115, line 53: Can you add the new argument to the docstring below?

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -133,10 +144,11 @@ def map_helper(self,
...
         Args:
           cur_node: node to examine.
-          mapping: current mapping of named qubits to `GridQubits`
-          available_qubits: current set of unassigned qubits
+          mapping: current mapping of named qubits to `GridQubits`,
+          available_qubits: current set of unassigned qubits,
           graph: adjacency graph of connections between qubits,
-            representing by a dictionary from qubit to adjacent qubits
+            representing by a dictionary from qubit to adjacent qubits,
+          nodes_trying: saves the current nodes under trying in this stack.
```
dstrain115, line 151: This is awkward wording:  "saves the current nodes under trying in this stack"

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -181,27 +193,32 @@ def map_helper(self,
...
         node_to_map = nodes_to_map[0]
 
+
         for node_to_try in valid_adjacent_qubits:
             # Add proposed qubit to the mapping
             # and remove it from available qubits
             mapping[node_to_map] = node_to_try
             available_qubits.remove(node_to_try)
-
+            nodes_trying.append(node_to_map)
```
dstrain115, line 202: Isn't this just the inverse of available_qubits?  Why do we need the extra variable?

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -181,27 +193,32 @@ def map_helper(self,
...
             if success:
                 success = self.map_helper(cur_node, mapping, available_qubits,
-                                          graph)
+                                          graph, nodes_trying)
 
             if not success:
                 # We have failed.  Undo this mapping and try another one.
-                del mapping[node_to_map]
-                available_qubits.add(node_to_try)
+                while True:
```
dstrain115, line 216: Should this check if nodes_trying is not empty instead of just being True?

recirq/quantum_chess/circuit_transformer_test.py
```diff
@@ -29,7 +29,60 @@
 c3 = cirq.NamedQubit('c3')
 d1 = cirq.NamedQubit('d1')
 
+@pytest.mark.parametrize('device',
```
dstrain115, line 32: Is there a particular test case where map_helper failed?  We should add it as a regression test.  It will also make it more clear why we need to adjust the search function.

recirq/quantum_chess/bit_utils.py
madcpf: Hi Doug, yes this is strange. I merged that PR first, then created this one. Not sure why this is still showing up.

recirq/quantum_chess/circuit_transformer.py
madcpf: Yes that's been added.

recirq/quantum_chess/circuit_transformer.py
madcpf: Now it's been re-phrased.

recirq/quantum_chess/circuit_transformer.py
madcpf: In a typical DFS we don't need this, but what's special here is:

# call this the 1st run
success = self.map_helper(node_to_map, mapping, available_qubits,
                                      graph, nodes_trying)
# call this the 2nd run
if success:
                success = self.map_helper(cur_node, mapping, available_qubits,
                                          graph, nodes_trying)

When the first run returns success, but the second run returns failure, according to the logic before this PR only the top-level nodes (i.e. node_to_try and node_to_map) will be released from mapping and added back into the available_qubits, but actually the nodes involved in the first call also need to released/added back.

The last test in circuit_transformer_test.py is a good example of this case.

recirq/quantum_chess/bit_utils.py
madcpf: Yes that's strange. I merged the last PR before committing this one for review. Not sure why this happens.

recirq/quantum_chess/circuit_transformer.py
madcpf: Thanks it's added now.

recirq/quantum_chess/circuit_transformer.py
madcpf: Thanks that's re-phrased.

recirq/quantum_chess/circuit_transformer.py
madcpf: In a normal DFS we don't need this extra variable. But what's special here is that 

# call this the 1st run
            success = self.map_helper(node_to_map, mapping, available_qubits,
                                      graph, nodes_trying)
# call this the 2nd run
            if success:
                success = self.map_helper(cur_node, mapping, available_qubits,
                                          graph, nodes_trying)

When the 1st run returns success but the second run returns failure, all nodes involved in 1st and 2nd run are supposed to be reversed in mapping/available_qubits, while before this PR the behavior is to only reverse the top-level ones (i.e. node_to_map, node_to_try). The result is that the loop after this will have less available_qubits to try for, and may end up with a failure in creating the mapping. One perfect example is the last testing case in circuit_transformer_test.py.

recirq/quantum_chess/circuit_transformer.py
madcpf: named_qubit == node_to_map will always be satisfied by one named_qubit in nodes_trying, so I think the current condition is fine.

recirq/quantum_chess/circuit_transformer_test.py
madcpf: Hi Doug, the last test in this file covers the situation corresponding to the map_helper() bug. That test will fail the bug is revealed if the fix is not been applied.

madcpf:
To all reviewers: The changes are only in circuit_transformer.py and circuit_transformer_test.py. All other changes have been reviewed and merged. I'm not sure why they still show up here, I tried to search for a resolution. One suggests git rebase, but that did not work. If you happen to know how to fix this, please let me know. Thanks a lot for your time reviewing this!

recirq/quantum_chess/circuit_transformer.py
dstrain115: Would it be possible to document this or explain it somewhere?  (Either in a comment/docstring or in the PR description)?

recirq/quantum_chess/bit_utils.py
dstrain115: I would try merging from master again.  The last merge from master was from 23 days ago, so that's probably why.

You should be able to do a 'git pull upstream master' in order to do this.

Can you do that before I review again?  Having all the extra stuff makes it more confusing to review.

madcpf: Thanks Doug. Now it's merged. Sorry for the confusion.

recirq/quantum_chess/circuit_transformer.py
madcpf: Thanks Doug. It's both added in the PR description and the docstring.

dstrain115 (approved):

recirq/quantum_chess/circuit_transformer_test.py
weinstein: the last test that you added? or the last test that was already in the file?

I'm just curious why circuit_transformer_test was not failing previously / is not failing without changes from this pull request -- if it were it should have stopped us from merging previous pull requests?

madcpf: The last test that was already in the file. Its debug output for device0 is:

**Before change:**

a1 -> (6, 3)
a2 -> (6, 4)
a3 -> (6, 5)
a4 -> (7, 5)
Not adjacent a1 and a4
Warning: could not map all qubits!
a3 -> (7, 4)
a4 -> (7, 5)
Not adjacent a1 and a4
a4 -> (7, 3)
a3 -> (7, 4)
a2 -> (6, 4)
b3 -> (6, 5)
a2 -> (6, 4)
a1 -> (6, 3)
b1 -> (6, 2)
c1 -> (6, 1)
d1 -> (5, 1)
Not adjacent a1 and d1
Warning: could not map all qubits!
c1 -> (7, 2)
Cannot fit adjacent nodes into c1
c1 -> (5, 2)
d1 -> (5, 3)
c1 -> (5, 2)
b1 -> (6, 2)
c2 -> (6, 1)
b1 -> (6, 2)
a1 -> (6, 3)

So it seems that even though before changing it starts with a wrong grid qubit, it ends up getting an accepted solution. But **after the top 2 DFS fixes** the problem shows up:

        for node_to_try in valid_adjacent_qubits:
            # Add proposed qubit to the mapping
            # and remove it from available qubits
            mapping[node_to_map] = node_to_try
>           available_qubits.remove(node_to_try)
E           KeyError: cirq.GridQubit(6, 3)

circuit_transformer.py:204: KeyError
---------------------------------------------------------- Captured stdout call -----------------------------------------------------------
a1 -> (5, 3)
a2 -> (5, 4)
Cannot fit adjacent nodes into a2
a2 -> (5, 2)
a3 -> (5, 1)
a4 -> (5, 0)
Not adjacent a1 and a4
a4 -> (6, 1)
Not adjacent a1 and a4
a4 -> (4, 1)
Not adjacent a1 and a4
Warning: could not map all qubits!
a3 -> (6, 2)
a4 -> (6, 3)
a3 -> (6, 2)
a2 -> (5, 2)
b3 -> (5, 1)
a2 -> (5, 2)
a1 -> (5, 3)
b1 -> (5, 4)
Cannot fit adjacent nodes into b1
b1 -> (4, 3)
Cannot fit adjacent nodes into b1
Warning: could not map all qubits!

**After applying all 3 fixes:**

a1 -> (5, 3)
a2 -> (5, 4)
Cannot fit adjacent nodes into a2
a2 -> (5, 2)
a3 -> (5, 1)
a4 -> (5, 0)
Not adjacent a1 and a4
a4 -> (6, 1)
Not adjacent a1 and a4
a4 -> (4, 1)
Not adjacent a1 and a4
Warning: could not map all qubits!
a3 -> (6, 2)
a4 -> (6, 3)
a3 -> (6, 2)
a2 -> (5, 2)
b3 -> (5, 1)
a2 -> (5, 2)
a1 -> (5, 3)
b1 -> (5, 4)
Cannot fit adjacent nodes into b1
b1 -> (4, 3)
Cannot fit adjacent nodes into b1
Warning: could not map all qubits!
a2 -> (6, 3)
a3 -> (6, 4)
a4 -> (6, 5)
Not adjacent a1 and a4
a4 -> (7, 4)
Not adjacent a1 and a4
a4 -> (5, 4)
a3 -> (6, 4)
a2 -> (6, 3)
b3 -> (6, 2)
a2 -> (6, 3)
a1 -> (5, 3)
b1 -> (5, 2)
c1 -> (5, 1)
d1 -> (5, 0)
Not adjacent a1 and d1
d1 -> (6, 1)
Not adjacent a1 and d1
d1 -> (4, 1)
Not adjacent a1 and d1
Warning: could not map all qubits!
c1 -> (4, 2)
d1 -> (4, 3)
c1 -> (4, 2)
b1 -> (5, 2)
c2 -> (5, 1)
b1 -> (5, 2)
a1 -> (5, 3)

weinstein: I see. So to Doug's point, can we come up with a test case that would have failed before the fixes (and without getting lucky) but passes now? That will help prevent future regressions if this code is refactored again in the future.

cantwellc (approved):

PR #162, opened 2021-04-06
Add a script for evaluating CircuitTransformer implementations and make some performance optimizations

weinstein:
* add circuit_transform_benchmark.py script for evaluating hardware circuit placement on generic (not necessarily chess) qasm circuits (ex: the ones from [the paper](https://ieeexplore.ieee.org/abstract/document/8976109) which I found [here](https://github.com/BensonZhou1991/circuittransform))
* make some profile-driven performance optimizations:
  - replace floyd-warshall with bfs when calculating graph centers during initial placement
  - more efficient implementation for generate_decomposed_swap in the swap updater
  - precompute device qubit neighbors in the swap updater
  - cache the active gates set in DependencyLists instead of recomputing them when they're accessed

relevant for #127 and #133


recirq/quantum_chess/experiments/circuit_transform_benchmark.py
```diff
@@ -0,0 +1,129 @@
...
+import pstats
+import sys
+
+import recirq.quantum_chess.circuit_transformer as ct
+from cirq.contrib.qasm_import import circuit_from_qasm
+from typing import List
+from timeit import default_timer as timer
+
+
+def print_stats(time_sec: float, circuit: cirq.Circuit) -> None:
```
dstrain115, line 36: Considering adding a short docstring to this function and the functions below.

recirq/quantum_chess/mcpe_utils.py
```diff
@@ -102,14 +102,39 @@ def __init__(self, circuit: cirq.Circuit):
...
+        for q in self.dependencies:
+            self._add_front_if_active(q)
+
+    def _add_front_if_active(self, qubit: cirq.Qid) -> None:
+        """Adds a qubit's front gate to self.active_gates if it is active."""
+        if not self.dependencies[qubit]:
+            return
+        # Get the gate at the front of the dependency list for this qubit.
+        # It's an active gate iff it is at the front of all the dependency lists
+        # of all qubits it operates on.
```
dstrain115, line 115: Optional: It may be worth moving this to the docstring rather than have it be line comments.  That would help with additional visibility if people want to understand what a front gate is (like in pycharm or in recirq docs)

recirq/quantum_chess/experiments/circuit_transform_benchmark.py
```diff
@@ -0,0 +1,129 @@
...
+  3) apply the CircuitTransformer
+  4) validate the resulting circuit for the Sycamore chip
+
+Some statistics about the transformation in step (3) will be printed, including
+timing and increases to the circuit size.
+
+Any qasm format circuit file should work.
+See https://github.com/BensonZhou1991/circuittransform for some examples which
+are commonly used in publications.
+"""
```
dstrain115, line 23: Is it worth adding an example qasm file to the repo, especially if there is a good example from the paper?

dstrain115 (commented):
I would update the PR description to explain that you are changing out Floyd-Warshall for BFS since this seems like a big enough change to mention in the commit.

recirq/quantum_chess/experiments/circuit_transform_benchmark.py
weinstein: Sure. Is there any reason we shouldn't just add all the ones referenced in the paper?

recirq/quantum_chess/mcpe_utils.py
weinstein: Sounds good. Updated the docstring.

recirq/quantum_chess/experiments/circuit_transform_benchmark.py
weinstein: Updated the docstrings in this file.

wingers (approved):

recirq/quantum_chess/experiments/circuit_transform_benchmark.py
dstrain115: Seems fine to me, assuming there are no copyright/license issues.

recirq/quantum_chess/experiments/circuits/4gt13_92.qasm
```diff
@@ -0,0 +1,70 @@
+OPENQASM 2.0;
```
dstrain115, line 1: Can we add a README.md to this directory to explain what all of these are?  I am not sure I can tell what 4gt13_92 means.

dstrain115 (approved):
LGTM once we add some description of the qasm files.

recirq/quantum_chess/experiments/circuits/4gt13_92.qasm
weinstein: It's not well documented anywhere I can find, and I'm not much for guessing names or reverse engineering qasm either :) I think what's most important is the size of the circuits so I made sure to document that

Issue #165, opened 2021-04-26
fermi_hubbard publication_test is not hermetic


weinstein:
it downloads datasets from https://datadryad.org/stash/dataset/doi:10.5061/dryad.crjdfn32v
the download URLs are not super stable, but also datadryad.org website unavailability can cause the recirq test suites to flake out and fail

It may be better to avoid the external dependency on datadryad.org in tests and mock out the download.

rmlarose:
Documenting discussion with the helpful folks at Dryad:

> Our team has confirmed that individual files do not have their own unique link. Previously, the did but that was changed with the relaunch in 2019. 

So a permanent fix should use the DOI of the dataset instead of URLs for individual files.

hiSandog:
Unit tests should use a small checked-in or generated fixture and mock the downloader so normal CI is hermetic. A separate, non-blocking integration job can resolve the dataset DOI, verify a pinned checksum/expected files, and alert when the upstream deposit changes. This keeps publication reproducibility tied to the DOI while preventing Dryad availability or URL layout from making unrelated test suites flaky.

PR #166, opened 2021-04-26
fix fermi_hubbard/publication_test by updating datadryad download URLs since they appear to have changed

weinstein:
See issue #165


recirq/fermi_hubbard/publication.py
```diff
@@ -241,10 +241,10 @@ def fetch_publication_data(
...
     base_url = "https://datadryad.org/stash/downloads/file_stream/"
     data = {
-        "gaussians_1u1d_nofloquet": "706210",
-        "gaussians_1u1d": "706211",
-        "trapping_2u2d": "706212",
-        "trapping_3u3d": "706213"
+        "gaussians_1u1d_nofloquet": "451326",
+        "gaussians_1u1d": "451327",
+        "trapping_2u2d": "451328",
+        "trapping_3u3d": "451329"
```
rmlarose, line 247: It's a bummer these aren't permanent identifiers. I reached out to dryad help and they essentially said the only thing permanent is the dataset DOI. So a long-term fix will probably need to locate files from the DOI instead of from /stash/downloads/file_stream.

mrwojtek:
@rmlarose , should this be merged and you can follow-up on the bug itself?

rmlarose:
> @rmlarose , should this be merged and you can follow-up on the bug itself?

That seems reasonable to me.

mrwojtek (approved):

Issue #152, opened 2021-03-02
Implement forced measurement outcome post-selection


cantwellc:
Measurements in Quantum Chess occur during excluded or capture moves. The current Cirq implementation runs 100 samples to determine the measurement outcome when it encounters one of those move variants, then stores that outcome as part of the move sequence history. Future sampling runs are post-selected based on that outcome. 

In order to support features like undo move, sandbox mode demonstrations, and game replays, we need to allow the user to provide a desired measurement outcome through the REST API (currently implemented separately as a server that forwards calls to this Cirq implementation). The Cirq implementation of Quantum Chess should then use this value for the post-selection, instead of the random outcome from the 100-shot described above.

losos0:
It was fixed by #167. Putting `.m0` in the notation works.

PR #167, opened 2021-04-26
Allow measurement outcomes to be provided to the quantum board when performing moves

weinstein:
Part of fixing issue #152
* adds an optional measurement outcome field to Move objects
* use that measurement outcome instead of sampling if present when performing moves

### DIFF
--- a/recirq/quantum_chess/move.py
+++ b/recirq/quantum_chess/move.py
@@ -46,7 +46,6 @@ class Move:
     have a move type and variant that determines what kind of move this is
     (capture, exclusion, etc).
     """
-
     def __init__(self,
                  source: str,
                  target: str,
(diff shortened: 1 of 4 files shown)


cantwellc (approved):
This should work nicely.

recirq/quantum_chess/move.py
```diff
@@ -46,35 +46,38 @@ class Move:
...
     def __init__(self,
                  source: str,
                  target: str,
                  *,
                  source2: str = None,
                  target2: str = None,
                  move_type: enums.MoveType = None,
-                 move_variant: enums.MoveVariant = None):
+                 move_variant: enums.MoveVariant = None,
+                 measurement: bool = None):
```
lingxz, line 57: Should we use an int for this instead? Is any one measurement "truer" than the other? And we also won't have to deal with the conversion in post_select later on.

cantwellc: That's a good thought. Int would also allow for the possibility of  multi outcome measurements on the future

PR #170, opened 2021-05-05
change Move.from_string format to use quantum chess algebraic notation

weinstein:
this makes at least the sources/targets part of the move strings consistent with the quantum chess game frontend, publication, and output string format of Move.

Also fix Move string parsing for split/merge moves when measurements are present which were broken by #167


lingxz (commented):
https://github.com/quantumlib/ReCirq/blob/master/recirq/quantum_chess/experiments/example_moves.txt this file should be changed too?

cantwellc (approved):
LGTM. This will work for the current data we have. I still have to fix the engine output to the database for the history, to include the extra information in move strings (like piece and variant), so we will likely need another update then. No definitive time frame on that though.

PR #171, opened 2021-05-10
Make FH `fetch_publication_data` more stable

### LINKED ISSUE #165: fermi_hubbard publication_test is not hermetic

rmlarose:
Fixes #165.

As documented there, individual file IDs are not permanent on Dryad. The strategy here is to determine these from the (permanent) DOI. 

Could be a cleaner way to do this, comments welcome.

### DIFF
--- a/recirq/fermi_hubbard/publication.py
+++ b/recirq/fermi_hubbard/publication.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 """Data specific to experiment published in arXiv:2010.07965."""
 
-from io import BytesIO
 from copy import deepcopy
+from io import BytesIO
 import os
+import re
 from typing import Callable, List, Optional, Tuple
 from urllib.request import urlopen
 from zipfile import ZipFile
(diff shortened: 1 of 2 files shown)


recirq/fermi_hubbard/publication.py
```diff
@@ -239,23 +240,26 @@ def fetch_publication_data(
...
-        data = {path: key for path, key in data.items() if path not in exclude}
-
-    for path, key in data.items():
+    fnames = {
+        "gaussians_1u1d_nofloquet", "gaussians_1u1d", "trapping_2u2d", "trapping_3u3d"
+    }.difference(exclude)
+
+    # Determine file IDs. Note these are not permanent on Dryad.
+    file_ids = {}
+    for line in urlopen("https://doi.org/10.5061/dryad.crjdfn32v").readlines():
```
mpharrigan, line 249: is this parsing the human-readable html website? If so, this seems just as unstable. What if they add a newline between the filename.zip and the id?

rmlarose: Yes this would break it. It might be that this is more stable than the file IDs - the file IDs changed at least once since `fetch_publication_data` was added, but I don't think the html has. I think the only way to hermetically seal it is to tell Dryad to give individual files DOIs, or use a different service that does.

mpharrigan:
see #352

PR #172, opened 2021-05-17
Add rep rate benchmark to ReCirq

dstrain115:
Adds a new module for benchmarks in ReCirq in which the only one is currently the rep rate calculator.

This class will take parameters specifying the size of the circuit (num_sweeps, width, depth, etc) and measure the latency of the circuit when sending to the service.  This will create repetition rate statistics (number of repetitions executed per seconds).

This class is used to generate repetition rate statistics for the processor data sheets.

### DIFF
--- a/recirq/benchmarks/rep_rate/rep_rate_calculator.py
+++ b/recirq/benchmarks/rep_rate/rep_rate_calculator.py
@@ -1,0 +1,227 @@
+"""
+  Class to calculate latency and repetition rate.
+
+  This class should replicate the acquisition of date for
+  rep rate and latency for processor data sheets.
+"""
+import random
+import cirq
+import cirq_google as cg
+import time
+import sympy
+
+from typing import List, Optional, Tuple
+
+
+class RepRateCalculator:
+    """Class to perform simple timing experiments using hardware.
+
+    This class calculates the repetition rate for various
+    types of circuits.  This is the approximate type per repetition
+    (or shot) to execute this on the device.  This rate varies widely
+    based on width, depth, and amount of sweeps/batching in the
+    circuit.  This class will help test various rates on the device.
+
+    This class is used to generate statistics on the processor datasheets
+    for latency and repetition rates.
+    """
+
+    def __init__(self,
+                 project_id: str,
+                 processor_id: str,
+                 gate_set: Optional[cg.SerializableGateSet] = None,
+                 device: Optional[cirq.Device] = None,
+                 sampler: Optional[cirq.Sampler] = None):
+        """Initialize the tester object.
+
+        Args:
+            project_id: Cloud project id as a string
+            processor_id: Processor to test on
+            gate_set: gate set to use.  Defaults to square root of iswap
+            device: device object to use. defaults to device provided by engine
+            sampler: sampler object to use.  defaults to engine sampler
+        """
+        if not device or not sampler:
+            engine = cg.Engine(project_id=project_id)
+        gate_set = gate_set or cg.SQRT_ISWAP_GATESET
+        self.device = device or engine.get_processor(processor_id).get_device(
+            gate_sets=[gate_set])
+        self.sampler = sampler or engine.sampler(processor_id=processor_id,
+                                                 gate_set=gate_set)
+        self.gate = cirq.ISWAP**0.5
+        self.qubits = list(self.device.qubits)
+        if gate_set == cg.SYC_GATESET:
+            self.gate = cirq.google.SYC
+
+        # log of print statements for testing
+        self.print_log = ''
+
+    def _flush_print_log(self):
+        self.print_log = ''
+
+    def _print(self, s:str):
+        """Includes printing to testing log and flushing."""
+        self.print_log += s
+        print(s, flush=True)
+
+    def _latency_circuit(self) -> cirq.Circuit:
+        """Circuit for measuring round-trip latency.
+
+        Use the most simple circuit possible, a measurement of a single
+        random qubit.
+        """
+        return cirq.Circuit(cirq.measure(random.choice(self.qubits)))
+
+    def _entangling_layer(self, qubits: List[cirq.Qid]) -> cirq.Moment:
+        """Creates a layer of random two-qubit gates."""
+        m = cirq.Moment()
+        for q in qubits:
+            if q in m.qubits:
+              continue
+            pairings = [
+                q + adj
+                for adj in [(0, 1), (1, 0), (0, -1), (-1, 0)]
+                if q + adj in qubits and q + adj not in m.qubits
+            ]
+            if not pairings:
+                continue
+            q2 = random.choice(pairings)
+            m=m.with_operation(self.gate(q, q2))
+        return m
+
+    def _sq_layer(self, qubits: List[cirq.Qid], parameterized: bool,
+                 symbol_start: int) -> (cirq.Moment, int):
+        """Creates a layer of single-qubit gates.
+
+        If parameteried is true, this will add symbols to the qubits
+        in order to test parameter resolution.
+        """
+        m = cirq.Moment()
+        current_sym = symbol_start
+        for q in qubits:
+            if parameterized:
+                symbol = f's_{current_sym}'
+                current_sym += 1
+                m=m.with_operation(cirq.X(q)**sympy.Symbol(symbol))
+            else:
+                m=m.with_operation(
+                    cirq.PhasedXZGate(x_exponent=random.random(),
+                                      z_exponent=random.random(),
+                                      axis_phase_exponent=random.random())(q))
+        return (m, current_sym)
+
+    def _create_rep_rate_circuit(self, parameterized: bool, width: int,
+                                depth: int, num_sweeps: Optional[int]) -> Tuple[cirq.Circuit, cirq.Sweep]:
+        """Creates a testing circuit based on parameters.
+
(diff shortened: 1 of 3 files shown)


maffoo:
Should this go in cirq_google? Seems like it might be nice to have this co-located with the rest of the engine code.

dstrain115:
> Should this go in cirq_google? Seems like it might be nice to have this co-located with the rest of the engine code.

Adding @mpharrigan for comments.
We were thinking about having a set of benchmarks for a feature testbed in reCirq, of which rep rate tester would be one of them.  This could also go in cirq_google if we decide that it should be separate from the other micro-benchmarks.  Matt H, do you have an opinion?

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+    (or shot) to execute this on the device.  This rate varies widely
+    based on width, depth, and amount of sweeps/batching in the
+    circuit.  This class will help test various rates on the device.
+
+    This class is used to generate statistics on the processor datasheets
+    for latency and repetition rates.
+    """
+
+    def __init__(self,
+                 project_id: str,
```
mpharrigan, line 30: consider making this optional and using `cg.get_engine` which will use the google cloud environment variable to get the project id if one is not specified

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+                 processor_id: str,
+                 gate_set: Optional[cg.SerializableGateSet] = None,
+                 device: Optional[cirq.Device] = None,
```
mpharrigan, line 33: Consider having the default constructor take explicit, required `(device, sampler, qubits, gate)` with a factory function (`classmethod`) which takes a processor_id and constructs using engine. I see `self.engine` is not saved as an attribute to this class.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+                 sampler: Optional[cirq.Sampler] = None):
```
mpharrigan, line 34: likewise: why would you want to override this

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+        self.gate = cirq.ISWAP**0.5
+        self.qubits = list(self.device.qubits)
+        if gate_set == cg.SYC_GATESET:
+            self.gate = cirq.google.SYC
+
+        # log of print statements for testing
+        self.print_log = ''
+
+    def _flush_print_log(self) -> None:
+        """Remove print log used fro debugging."""
```
mpharrigan, line 60: fro -> for

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+        self.print_log = ''
+
+    def _print(self, s: str) -> None:
+        """Includes printing to testing log and flushing."""
+        self.print_log += s
+        self.print_log += '\n'
+        print(s, flush=True)
+
+    def _latency_circuit(self) -> cirq.Circuit:
```
mpharrigan, line 69: I'm going to make a lot of comments about basically the same thing. If you want to keep the Java-style methods let me know but:

this function can (almost) be a free function. It only uses `self.qubits`.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+        """Circuit for measuring round-trip latency.
+
+        Use the most simple circuit possible, a measurement of a single
+        random qubit.
+        """
+        return cirq.Circuit(cirq.measure(random.choice(self.qubits)))
+
+    def _entangling_layer(self, qubits: List[cirq.Qid]) -> cirq.Moment:
```
mpharrigan, line 77: This can almost be a free function. It only uses `self.gate`. Why does this not use `self.qubits`?

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+                for adj in [(0, 1), (1, 0), (0, -1), (-1, 0)]
+                if q + adj in qubits and q + adj not in m.qubits
+            ]
+            if not pairings:
+                continue
+            q2 = random.choice(pairings)
+            m = m.with_operation(self.gate(q, q2))
+        return m
+
+    def _sq_layer(self, qubits: List[cirq.Qid], parameterized: bool,
```
mpharrigan, line 94: this can be a free function or should be annotated with `@staticmethod`

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+                current_sym += 1
+                m = m.with_operation(cirq.X(q)**sympy.Symbol(symbol))
+            else:
+                m = m.with_operation(
+                    cirq.PhasedXZGate(x_exponent=random.random(),
+                                      z_exponent=random.random(),
+                                      axis_phase_exponent=random.random())(q))
+        return (m, current_sym)
+
+    def _create_rep_rate_circuit(
```
mpharrigan, line 115: this can almost be a free function

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+        The circuit will be of alternating single and two qubit layers
+        using a number of qubits specified by width and a number of
+        moments specified by depth.
+
+        This function will also create a sweep if parameterized is true,
+        by parameterized all single qubit layers with random values for
+        angles.
+        """
+        qubits = self.qubits[:width]
+        c = cirq.Circuit()
```
mpharrigan, line 129: would `cirq.experiments.random_quantum_circuit_generation` functions work here?

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -0,0 +1,263 @@
...
+    def print_latency(self, num_trials: int) -> None:
+        self._flush_print_log()
+        latencies = self.get_latency_samples(num_trials)
+        latencies = sorted(latencies)
+        median = latencies[len(latencies) // 2]
+        p05 = latencies[len(latencies) // 20]
+        p95 = latencies[-len(latencies) // 20]
+        p10 = latencies[len(latencies) // 20]
+        p90 = latencies[-len(latencies) // 20]
+        average = sum(latencies) / len(latencies)
```
mpharrigan, line 178: might be less error prone and faster to use `np.median` `np.percentile`

mpharrigan:
> Should this go in cirq_google? Seems like it might be nice to have this co-located with the rest of the engine code.

I'd categorize this as "something you run" not "something you use to run things". You could maybe argue that it should go in `cirq.qcvv` as it *is* characterizing something but I don't think many peoples' first thought about qcvv is rep rate.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Restructured.  Let me know what you think.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Restructured.  I think you might want to override sampler for testing, simulating, or mocking though.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Yeah, I am not sure.  I punted on that for now.  The tricky part would be for parameterizing the circuits, which I don't think this library can do.  I would definitely like to keep parameter sweeps in this, since it is a crucial component of rep rate (and we have had problems with parameterizing being slow before).

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Done.

mpharrigan (approved):
sorry for letting this get lost

PR #175, opened 2021-06-10
Quantum Chess: Fix post-selection for empty squares

dstrain115:
For excluded moves, measurement outcomes means that the
move is successful.  The move is successful if the target
square was empty (so the piece could move there), so a
measurement of 1 means that we are post-selecting on the
target square to be |0>.


weinstein (approved):
Thanks, good catch.

PR #182, opened 2021-07-13
Update Cirq to 0.11 and Fixes

mpharrigan:
- PyTKet checks for gate equality instead of doing isinstance() checks, so Cirq's change to having a real Rx class breaks compilation.
 - PyTKet got rid of `Device`
 - get engine calibration moved
 - some (but not all) `cirq.google` -> `cirq_google` migration


dstrain115:
We should change the PR name and description to match the fact that multiple changes (wrap_function, nbfmt, etc) are involved in this PR.  Otherwise, it looks good so far.

mpharrigan:
I was using this PR as a little testbed. I hope to merge those other changes independently and then rebase this one so it's only the Rx change

recirq/qaoa/problem_circuits_test.py
```diff
@@ -42,38 +42,38 @@ def test_get_routed_grid_model_circuit():
...
+  │      │      │       │       │      │
+┌╴│      │      │       │       │      │     ╶┐
+│ ZZ─────┼──────┼───────ZZ      │      │      │
+│ │      ZZ─────┼───────┼───────ZZ     │      │
+│ │      │      ZZ──────┼───────┼──────ZZ     │
+└╴│      │      │       │       │      │     ╶┘
+  │      │      │       │       │      │
+  │      │      │       │       │      │
+  │      │      │       │       │      │
+  X      X      X       X       X      X
```
mpharrigan, line 61: annoyingly, this test behavior will now change depending on cirq version that you're using recirq with, although the circuits have the same effect.

mpharrigan:
We don't have an automated test for the notebooks, but the fermi hubbard example is currently broken by the loss of `cirq.TrialResult` backwards compatibility https://github.com/quantumlib/Cirq/issues/4318

recirq/qaoa/problem_circuits_test.py
dstrain115: Optional:  Is there a better way to test it besides testing the diagram output?  If not, oh well I guess?

dstrain115 (approved):

recirq/qaoa/problem_circuits_test.py
mpharrigan: I'll investigate. I think we have some tests for unitary equivalence, which is the important thing. This may have been an additional test to be sure (or to check for keeping the circuit structured how we expected)

requirements.txt
```diff
@@ -1,4 +1,5 @@
-cirq~=0.10.0
+cirq-core~=0.11.0
```
mpharrigan, line 1: todo: require 0.11.1

PR #183, opened 2021-07-13
Install tensorflow-docs requirements for notebook check

mpharrigan:
The check seemed to be failing on another PR. This fixes it. cc @rmlarose who wrote the original downloader script.

PR #184, opened 2021-07-13
Deal with scipy unceremoniously removing `wrap_function`

mpharrigan:


mpharrigan:
This fixes the issue seen in https://github.com/quantumlib/ReCirq/pull/183 and https://github.com/quantumlib/ReCirq/pull/183 fixes the issue seen in this

madcpf:
Thanks for fixing this!

balopat (approved):
LGTM  

For the record, it is a verbatim copy from https://github.com/scipy/scipy/blob/maintenance/1.6.x/scipy/optimize/optimize.py#L454. 

Consider adding the https://github.com/scipy/scipy/blob/maintenance/1.6.x/LICENSE.txt to the top of this file referring to wrap_function.

mpharrigan:
added

recirq/optimize/__init__.py
```diff
@@ -12,4 +12,53 @@
...
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+# SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+def wrap_function(function, args):
```
dstrain115, line 49: I generally don't think it's a great idea to put code in __init__ files.  Is there a reason we can't put this in its own file and import it into the package?

recirq/optimize/__init__.py
```diff
@@ -12,4 +12,53 @@
...
+    # This function existed in scipy prior to 1.7
+    # It must not have been considered part of the public API, as it has
+    # disappeared without a trace.
```
dstrain115, line 52: Should this be a docstring instead of line comments?

dstrain115 (approved):

recirq/optimize/__init__.py
mpharrigan: done

recirq/optimize/__init__.py
mpharrigan: done, moved

balopat (approved):

PR #185, opened 2021-07-14
Format quantum chess notebooks, fixing mixed indentation levels and recirq reinstall

### LINKED ISSUE #180: Mixed indentation levels in Quantum Chess notebooks
concepts and rest api notebook sometimes use 2 spaces per indent level, when everywhere else 4 is used.

### LINKED ISSUE #181: quantum_chess_client.ipynb reinstalls recirq
docs/quantum_chess/quantum_chess_client.ipynb does not do the normal thing of trying to `import recirq` and only installing if it's not found. Otherwise, it will overwrite a developer's install.

rmlarose:
Fixes #180. I just ran [black-nb](https://pypi.org/project/black-nb/) on the notebooks. This fixes the indentation level but also introduces some other changes, e.g. double quotes instead of single quotes for strings. (Sorry to jump in if Doug *really* wanted to take this issue.)

Also checks if recirq is already installed to fix #181.


docs/quantum_chess/concepts.ipynb
```diff
@@ -18,7 +18,7 @@
    },
    "outputs": [],
    "source": [
-    "#@title Licensed under the Apache License, Version 2.0 (the \"License\");\n",
+    "# @title Licensed under the Apache License, Version 2.0 (the \"License\");\n",
```
mpharrigan, line 21: will this break the devsite parsing?

mpharrigan:
Thanks! @dstrain115 can you take a look or have someone from qchess take a look?

mpharrigan:
Test failure should be fixed by merging master

docs/quantum_chess/concepts.ipynb
rmlarose: I don't think so but other notebooks don't have a space, so changed to that.

dstrain115 (approved):
Looks like just formatting changes, so no issues that I can see.

As a side note, are we standardizing on black now instead of yapf for recirq?  Some of the quantum chess people were having formatting issues, and I think they were still using yapf.

mpharrigan:
There's no enforced formatting check in the CI. I personally use the PyCharm formatter. I'd recommend `black` over `yapf` since that's what Cirq uses.

PR #186, opened 2021-07-17
Add caching for moves, starting with SPLIT_JUMP

lingxz:
- Currently only caches SPLIT_JUMP from full square to 2 empty squares
- cache_results is not called anywhere yet, so it doesn't change existing flows


losos0:
Is there some design for how this will eventually work at a high level?
Like what is the scope of the cached data - e.g. in one QuantumBoard object, multiple boards in the same process?
How to decide if a position matches something in a cache?
How will runs of circuits for the purpose of caching be triggered?

lingxz:
@losos0  Sorry I missed your comment! Yes there is a design that where Chris outlined some of the potential things we can cache: https://docs.google.com/document/d/1mVkV18e5JRzY2n1qyB06kUbVGdwVjyY4H8IzkX36hJ4/edit?usp=sharing. 

> How to decide if a position matches something in a cache?
I think this will depend on the kind of move that was cached. I think the general idea will be to determine what kind of move it is (split jump, or jump from a previously split square to an empty target, etc. which should be encapsulated by the CacheKey dataclass), then check if this move matches anything in the cache. If we do something more complex, then it would just require specifying more things in CacheKey to indicate which cached results it can match.

recirq/quantum_chess/quantum_board.py
```diff
@@ -331,16 +334,81 @@ def _generate_accumulations(self, repetitions: int = 1000) -> None:
...
     def get_probability_distribution(self,
-                                     repetitions: int = 1000) -> List[float]:
+                                     repetitions: int = 1000,
+                                     use_cache=True) -> List[float]:
         """Returns the probability of a piece being in each square.
 
         The values are returned as a list in the same ordering as a
         bitboard.
         """
+        move_history_cache_key = (tuple(self.move_history), repetitions)
```
losos0, line 387: As it stands `get_full_squares_bitboard` won't be in sync with the cached results and will trigger a new run of the circuit. Maybe this stuff should be in `_generate_accumulations` so that the full/empty bitboard stuff gets updated too?

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -964,6 +965,34 @@ def test_caching_accumulations_same_repetition_cached(board):
...
+                   move_variant=enums.MoveVariant.BASIC)
+    b.do_move(m1)
+    probs = b.get_probability_distribution(100)
+    b.do_move(m2)
+    # Expected probability with the cache applied
+    probs[square_to_bit('b1')] = 0
+    probs[square_to_bit('c1')] = b.cache[cache_key]["target"]
+    probs[square_to_bit('d1')] = b.cache[cache_key]["target2"]
+    # Get probability distribution should apply the cache without rerunning _generate_accumulations.
+    assert probs == b.get_probability_distribution(100, use_cache=True)
```
losos0, line 993: As a way to test that caching was used without accessing the internal state, how about using the debug log function to ensure that no additional circuits ran?

recirq/quantum_chess/quantum_board.py
```diff
@@ -331,16 +334,81 @@ def _generate_accumulations(self, repetitions: int = 1000) -> None:
...
+                self.probabilities = probs
+                self.move_history_probabilities_cache[
+                    move_history_cache_key] = probs.copy()
+                del self.cache[cache_key]
+                return probs
+
         if self.accumulations_repetitions != repetitions:
             self._generate_accumulations(repetitions)
 
+        self.move_history_probabilities_cache[
```
losos0, line 410: How about making this an array instead of having lists of moves as dictionary keys?

The array could be the same length as `move_history` and the entries a 2-tuple of (repetitions, probabilities) (with -1 for repetitions if the circuit was never run for that move).

Then you could just check if the entry at index -2 has enough repetitions. (IIUC, this PR is only implementing the "Generic Moves" variant as described in the doc so the move lists don't matter except to see whether we know the probabilities after the second-to-last move.)

recirq/quantum_chess/quantum_board.py
```diff
@@ -27,12 +27,16 @@
...
 )
+from recirq.quantum_chess.caching_utils import CacheKey, cache_key_from_move
 import recirq.quantum_chess.circuit_transformer as ct
 import recirq.quantum_chess.enums as enums
 import recirq.quantum_chess.move as move
 import recirq.quantum_chess.quantum_moves as qm
 
 
+# This is the basis state corresponding to FEN: rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR.
+DEFAULT_CHESS_INIT_STATE = 18446462598732906495
```
losos0, line 38: Hex could be more readable, `0xffff00000000ffff`

recirq/quantum_chess/quantum_board.py
```diff
@@ -331,19 +369,64 @@ def _generate_accumulations(self, repetitions: int = 1000) -> None:
 
         self.accumulations_repetitions = repetitions
 
+    def _caching_supported(self, m: move.Move):
+        """Checks if caching is supported for this move."""
+
+        # Caching is supported for a split jump from one full square to two empty squares.
+        if m.move_type == enums.MoveType.SPLIT_JUMP and nth_bit_of(square_to_bit(m.source), self.full_squares)\
```
losos0, line 376: self.full/empty squares are only available if a sample has been taken, which is not always the state. Instead use `self.entangled_squares` (which squares are not known to be in a classical state), and `self.state` which tracks occupancy for non-entangled squares. See [here](https://github.com/quantumlib/ReCirq/blob/4431f47bc0bf6f87a95adfbe0674026f8cd1506a/recirq/quantum_chess/quantum_board.py#L586) for an example.

recirq/quantum_chess/quantum_board.py
losos0: Yeah I think it's rare enough to ignore. In general I don't think the behavior of matching an exact number of repetitions makes sense; it would be better to use a cached value if it has *at least* as many samples as requested. And then we could store just the largest sample... that's out of scope for this PR of course :)

recirq/quantum_chess/quantum_board.py
```diff
@@ -311,10 +317,42 @@ def sample(self, num_samples: int) -> List[int]:
...
+                    self.cache[cache_key])
+                self.probabilities = probs
+                self._set_full_empty_squares_from_probability()
+                self.move_history_probabilities_cache[
+                    move_history_cache_key] = probs.copy()
+                del self.cache[cache_key]
+                return
+
         self.probabilities = [0] * 64
         self.full_squares = (1 << 64) - 1
```
losos0, line 357: Use the new full squares algorithm here also instead of having 2 implementations

recirq/quantum_chess/quantum_board.py
lingxz: Done

recirq/quantum_chess/quantum_board.py
lingxz: Checking entangled_squares doesn't seem to work because do_move is usually called before getting the probability distribution, and do_move already adds the targets in the split_jump to entangled squares. Do you think I can just set an initial value for full and empty squares?

recirq/quantum_chess/quantum_board.py
lingxz: Done.

recirq/quantum_chess/quantum_board.py
lingxz: Done.

recirq/quantum_chess/quantum_board.py
```diff
@@ -311,39 +323,123 @@ def sample(self, num_samples: int) -> List[int]:
...
         The values are returned as a list in the same ordering as a
         bitboard.
         """
         if self.accumulations_repetitions != repetitions:
-            self._generate_accumulations(repetitions)
+            self._generate_accumulations(repetitions, use_cache)
 
+        # Cache the repetiion and probability corresponding to this move history.
+        if len(self.move_history_probabilities_cache) < len(self.move_history):
+            self.move_history_probabilities_cache.extend([(-1, None)] * (len(self.move_history) - len(self.move_history_probabilities_cache)))
```
losos0, line 437: How about making the length of `self.move_history_probabilities_cache` always match the length of `self.move_history`? Just adding the dummy `(-1, None)` entry whenever the move history changes. Then there will won't need to be logic checking its length.

recirq/quantum_chess/move.py
```diff
@@ -72,6 +72,9 @@ def __eq__(self, other):
                     and self.measurement == other.measurement)
         return False
 
+    def __hash__(self):
```
losos0, line 75: No longer needed

lingxz: removed.

recirq/quantum_chess/quantum_board.py
lingxz: Thanks, done.

recirq/quantum_chess/quantum_board.py
lingxz: no I mean, the do_move already modifies `entangled_squares`, so we can't use it to check whether the target is empty or entangled. or am I misunderstanding?

losos0: Oh yes, you are right. I suppose you could work around this by either creating arrays of full/empty squares as known from `entangled_squares`/`state` updated after each move. Or alternatively, determine whether caching is supported inside of `do_move` and store that boolean somewhere.

recirq/quantum_chess/quantum_board.py
```diff
@@ -325,36 +337,113 @@ def sample(self, num_samples: int) -> List[int]:
...
+                    for i in range(64):
+                        # Assume initial state is classical
+                        previous_probability[i] = nth_bit_of(i, self.state)
+
+                probs = self._apply_cache(previous_probability, last_move,
+                    self.cache[cache_key])
+                self.probabilities = probs
+                self._set_full_empty_squares_from_probability()
+                self.move_history_probabilities_cache.append((repetitions, probs.copy()))
+                del self.cache[cache_key]
```
losos0, line 377: Comment on why the cache entry is deleted?

lingxz: Added a comment.

recirq/quantum_chess/quantum_board.py
lingxz: For all uses of the cache apart from the first move, we will need to have samples for the previous move anyway, since we need the prob distribution from the previous move to use the cache, so it won't be a problem. For the first move, I can just set the full_squares and empty_squares based on the init_basis_state (like what I did in this PR). It seems to work, do you see any problems?

recirq/quantum_chess/quantum_board.py
```diff
@@ -598,6 +689,7 @@ def do_move(self, m: move.Move) -> int:
 
         # Add move to move_history
         self.move_history.append(m)
+        self.move_history_probabilities_cache.append((-1, None))
```
losos0, line 692: move_history_probabilities_cache entry should be removed in undo_move

recirq/quantum_chess/quantum_board.py
losos0: OK, sounds good to me. Although the answer to whether a move should be cached could be "wrong" in principle, but the results would be off only a small amount and it would not change the circuit.

losos0 (approved):
LGTM besides the one undo_move comment

recirq/quantum_chess/quantum_board.py
lingxz: undo_move calls `with_state()`, which clears move_history_probabilities_cache and move_history

losos0: OK, then LGTM!

Issue #189, opened 2021-07-25
Unexpected keyword argument in `build_otoc_circuits`


rmlarose:
If you call `build_otoc_circuits` with `use_physical_cz=False`, `cz_to_sqrt_iswap` is called with `corrections=cz_correction`

https://github.com/quantumlib/ReCirq/blob/144323f212757775a8541dbe2f938b64534b25ec/recirq/otoc/otoc_circuits.py#L287

but `cz_to_sqrt_iswap` has no `corrections` argument

https://github.com/quantumlib/ReCirq/blob/144323f212757775a8541dbe2f938b64534b25ec/recirq/otoc/utils.py#L229-L232

To reproduce, run `python otoc_example.py` with [`use_physical_cz=False`](https://github.com/quantumlib/ReCirq/blob/master/recirq/otoc/otoc_example.py#L75).

```python
TypeError: cz_to_sqrt_iswap() got an unexpected keyword argument 'corrections'
```

PR #190, opened 2021-07-27
Add OTOC documentation

rmlarose:
Adds a tutorial based on [this example](https://github.com/quantumlib/ReCirq/blob/144323f212757775a8541dbe2f938b64534b25ec/recirq/otoc/otoc_example.py) which shows how to recreate Fig 1 from the [publication](https://arxiv.org/abs/2101.08870).


docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "D-6CzQFH84Tr"
+   },
+   "source": [
+    "Quantum scrambling describes how interaction in a quantum system disperses local information into its many degrees of freedom. Analogous to classical chaos, scrambling manifests itself as a [butterfly effect](https://en.wikipedia.org/wiki/Butterfly_effect) wherein a local perturbation is rapidly amplified over time. The perturbation is modeled by a local operator $\\hat{O}$ (the *butterfly operator*) acting on a qubit (the *butterfly qubit*). Under a dynamical process $\\hat{U} = \\hat{U}(t)$, the butterfly operator evolves in the Heisenberg picture as\n",
+    "\n",
+    "$$\n",
+    "\\hat{O} (t) = \\hat{U}^\\dagger \\hat{O} \\hat{U}\n",
```
mpharrigan, line 94: kindof strange that $t$ doesn't appear on the rhs

docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+    "\n",
+    "over a collection of quantum circuits with microscopic differences (e.g., in the phases of gates). Operator scrambling is then reflected in the average OTOC value $\\bar{C}(t)$ over different circuits:\n",
+    "\n",
+    "- $\\bar{C}(t) = 1$ when $\\hat{O}(t)$ and $\\hat{M}$ do not overlap.\n",
+    "- $\\bar{C}(t) < 1$ when $\\hat{O}(t)$ and $\\hat{M}$ overlap.\n",
+    "- $\\bar{C}(t) \\rightarrow 0$ in the fully scrambled limit where the commutation between $\\hat{O}(t)$ and $\\hat{M}$ is completely randomized.\n",
+    "\n",
+    "Note that if operator entanglement is also present ($n_p \\gg 1$), $\\bar{C}(t)$ approaches $0$ for all circuits and their fluctuation vanishes as well.\n",
+    "\n",
+    "In what follows, we show how to build OTOC circuits, run them & compute $\\bar{C}$, then visualize results and detect operator spreading."
```
mpharrigan, line 122: nit why do you use "&" once but "and" in other places

docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+    "            print(\"Notebook is not executed with Colab, assuming Application Default Credentials are setup.\") \n",
+    "\n",
+    "    authenticate_user()\n",
+    "    print(\"Successful authentication to Google Cloud.\")\n",
+    "\n",
+    "processor_id = \"\" #@param {type:\"string\"}\n",
+    "if use_noisy_simulator or processor_id == \"\":\n",
+    "    print(\"Using a noisy simulator.\")\n",
+    "    sampler = cg.PhasedFSimEngineSimulator.create_with_random_gaussian_sqrt_iswap(\n",
+    "        mean=cg.SQRT_ISWAP_PARAMETERS,\n",
```
mpharrigan, line 262: no such thing SQRT_ISWAP_PARAMETERS. Do you mean SQRT_ISWAP_INV_PARAMETERS

(non-optional)

docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {
+    "id": "GZIsJoNX1omO"
+   },
+   "outputs": [],
+   "source": [
+    "if not use_noisy_simulator:\n",
```
mpharrigan, line 308: `or processor_id == ""`

[or fix it so `use_noisy_simulator` is True if processor_id == ""]. (non optional)

docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+      "(7, 5): ───Y^0.5────────────PhX(-0.25)^0.5───@───PhX(-0.25)^0.5───────────────PhX(0.75)^0.5────@───PhX(0.75)^0.5──────────────────────────────\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"Ancilla qubit:\", ancilla)\n",
+    "print(\"Measurement qubit:\", qubits[0])\n",
+    "print(\"Butterfly qubit:\", q_b)\n",
+    "\n",
+    "print(\"\\nExample OTOC circuit:\\n\", otoc_circuits.butterfly_Y[0], sep=\"\\n\")"
```
mpharrigan, line 496: on my jupyter, whenever one does `print(circuit)` it wraps horribly and you can't tell what's going on but if you just have the last line of the cell be `circuit` it uses a horizontal scrollbar and I can actually read it

docs/otoc/otoc_example.ipynb
```diff
@@ -0,0 +1,761 @@
...
+   "metadata": {
+    "id": "zRsJMaG3EkgD"
+   },
+   "outputs": [],
+   "source": [
+    "# Compile to sqrt_iswap.\n",
+    "for i, circuits_trial in enumerate(circuit_list):\n",
+    "    for j, circuits_cycle in enumerate(circuits_trial):\n",
+    "        for k, circuit_otoc in enumerate(circuits_cycle):\n",
+    "            circuit_list[i][j][k] = cg.optimized_for_sycamore(circuit_otoc, optimizer_type=\"sqrt_iswap\")"
```
mpharrigan, line 600: this takes longer than expected on my machine. maybe `print('.', end='')` just so there's some indication of progress

mpharrigan: this actually takes quite a long time. maybe we can throw multiprocessing at it or have a real progress bar?

mpharrigan (changes requested):
I've noted the "non optional" comments, other comments should be taken under advisement

This is pretty slow, but I guess it comes with the territory. It's also rather unfortunate that we use "meaningful indexing" where you have to sortof note what the various indices mean instead of nice dataclasses, but I understand that's how the code is written and it's outside the scope of this example to change it.

mpharrigan:
should we remove the example script in favor of this notebook?

docs/otoc/otoc_example.ipynb
rmlarose: No change to match with arxiv paper

docs/otoc/otoc_example.ipynb
rmlarose: Changed & -> and

docs/otoc/otoc_example.ipynb
rmlarose: Fixed. (1st PR was before Cirq 0.11 update)

docs/otoc/otoc_example.ipynb
rmlarose: Fixed so `use_noisy_simulator` is True if `processor_id == ""`

docs/otoc/otoc_example.ipynb
rmlarose: Changed to just `circuit` instead of `print(circuit)`

docs/otoc/otoc_example.ipynb
rmlarose: Added a progress bar. Compiling is kind of a hack around #189 - ideally I think the desired behavior is that each circuit already has the right gates and no compiling is necessary.

rmlarose:
Thanks @mpharrigan! Ready for round 2.

> should we remove the example script in favor of this notebook?

Probably? There's nothing in the example script that isn't in the notebook.

docs/otoc/otoc_example.ipynb
mpharrigan: ok thanks for the additional information

mpharrigan (approved):

mpharrigan:
Thanks @rmlarose 

If you want to remove the example script, we can do that in a follow-up PR!

Issue #193, opened 2021-08-08
Ancillas are unsafe to use for computation


losos0:
For several kinds of quantum chess moves, ancilla qubits are used as scratch space for some computation. But this is in general not correct: if the ancilla's value depends on the board position (i.e. the overall state is not a product state of the new ancilla(s) and the rest of the qubits), then the phases between different positions will be messed up.

In this example, the knight should end up entirely on g1 at the end, but it does not merge correctly due to the d1-h5 path ancilla:
```
e2e4
g1^h3f3:SPLIT_JUMP:BASIC
f1^e2b5:SPLIT_JUMP:BASIC
d1h5:SLIDE:BASIC
d1h5:SLIDE:BASIC
d1h5:SLIDE:BASIC
d1h5:SLIDE:BASIC
h3f3^g1:MERGE_JUMP:BASIC
```

In this example, a merge slide also does not work due to the existence of path ancillas:
```
d2d4
c1^e3g5:SPLIT_SLIDE:BASIC
e3g5^c1:MERGE_SLIDE:BASIC
```

In the case of slide moves, it would be possible to zero out the path ancilla after the slide is performed, by repeating the operations used to create it. However, this could not be easily done in the case of merge slide, because there path qubits may coincide with source qubits, meaning the information used to create the ancilla is not easily accessible.

losos0:
Actually it is just a bug that, for the merge slide, the source squares are sometimes included in the path. The source squares should be excluded from the path qubits. If I fix that then it is easy to zero the ancillas after the slide. So maybe it will be possible to zero the ancillas in all cases.

losos0:
I believe these cases are fixed.

PR #194, opened 2021-08-12
Fix dirty ancillas for merge and slide moves

losos0:
Fixes some cases of #193. There are still other cases of ancilla usage left to check.

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -462,6 +462,14 @@ class CirqBoard:
             qm.controlled_operation(cirq.X, [path_ancilla], [], path_qubits))
         return path_ancilla
 
+    def _clear_path_ancilla(self, path_qubits, ancilla):
+        """Zeroes the path ancilla based on path_qubits.
+
+        This does the inverse operation of _create_path_ancilla.
+        """
+        self.circuit.append(
+            qm.controlled_operation(cirq.X, [ancilla], [], path_qubits))
+
     def set_castle(self, sbit: int, rook_sbit: int, tbit: int,
                    rook_tbit: int) -> None:
         """Adjusts classical bits for a castling operation."""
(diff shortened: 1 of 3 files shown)


recirq/quantum_chess/quantum_board.py
```diff
@@ -462,6 +462,10 @@ def create_path_ancilla(self, path_qubits: List[cirq.Qid]) -> cirq.Qid:
             qm.controlled_operation(cirq.X, [path_ancilla], [], path_qubits))
         return path_ancilla
 
+    def _clear_path_ancilla(self, path_qubits, ancilla):
```
dstrain115, line 465: It might be good to explain in a docstring what this is doing, since it is adding an operation to the circuit rather than explicitly "clearing" a structure.

recirq/quantum_chess/quantum_board.py
```diff
@@ -631,7 +635,11 @@ def do_move(self, m: move.Move) -> int:
 
             # Find all the squares on both paths
             path_qubits = self.path_qubits(m.source, m.target)
+            if tqubit2 in path_qubits:
```
dstrain115, line 638: Do you think we should fold this into the path_qubits() function (maybe have an optional exclude keyword argument?)

recirq/quantum_chess/quantum_board.py
```diff
@@ -644,6 +652,8 @@ def do_move(self, m: move.Move) -> int:
                 self.circuit.append(
                     qm.split_slide(squbit, tqubit, tqubit2, path1, path2,
                                    ancilla))
+                self._clear_path_ancilla(path_qubits, path1)
```
dstrain115, line 655: This is not part of the PR but we should probably keep underscores consistent between create_path_ancilla and _clear_path_aniclla.  Probably correct is to add the underscore in create_path_ancilla.

dstrain115 (approved):
Mainly just nits and minor comments.

recirq/quantum_chess/quantum_board.py
losos0: Done.

recirq/quantum_chess/quantum_board.py
losos0: Renamed to `_create_path_ancilla`.

recirq/quantum_chess/quantum_board.py
losos0: I considered it, but I prefer how it is now where path_qubits corresponds to a single clear concept.

PR #195, opened 2021-08-16
pin versions of tket

MichaelBroughton:
https://www.python.org/dev/peps/pep-0440/#compatible-release

It looks like in future versions this has been replaced with the BackendInfo class (https://cqcl.github.io/pytket/build/html/changelog.html?highlight=device)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,8 +9,8 @@ scikit-learn
 
 # qaoa
 networkx
-pytket~=0.8
-pytket-cirq~=0.8
+pytket~=0.8.0
+pytket-cirq~=0.8.0
 
 # quantum chess, only needed for tests
 scipy


mpharrigan:
I already made the change to not use pytket device. This should work on the newest pytket

mpharrigan:
Indeed this is failing the CI

MichaelBroughton:
There is a disparity between the required version of pytket here. In order for the library to work it needs the latest pytket, but in order for the `routing_with_tket.ipynb` notebook to work it needs 0.8.0. Which version should be pinned to ?

mpharrigan:
I'll submit a fix for the actual notebook

MichaelBroughton:
SGTM

PR #201, opened 2021-08-22
Fix deterministically flaky tests

losos0:
Currently the noisy simulator uses a random generator which is
    initialized to a fixed seed, but shared across tests. This means that if
    you run the same test command twice in a row you'll get the same
    results, but adding or modifying any test can affect the result of any
    other test. An annoying kind of deterministic flakiness. (The non-noisy
    simulators are currently using different seeds each time.)

This change makes all simulators use randomly chosen seeds by default.
    The seed can be forced to a specific value using the
    RECIRQ_CHESS_TEST_SEED environment variable. Seeds are written to the
    standard output so that when a test fails in pytest, you can see the
    seed and use it to reproduce a failure.


dstrain115 (approved):
Seems like a good solution to me.  Thanks for detecting and fixing!

PR #202, opened 2021-08-22
Fix two chess movement bugs

losos0:
- Split move incorrectly clears source square
- Pawn capture doesn't work right

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -848,8 +848,6 @@ class CirqBoard:
             tqubit2 = bit_to_qubit(tbit2)
             self.add_entangled(squbit, tqubit, tqubit2)
             self.circuit.append(qm.split_move(squbit, tqubit, tqubit2))
-            self.state = set_nth_bit(sbit, self.state, False)
-            self.unhook(squbit)
             return 1
 
         if m.move_type == enums.MoveType.MERGE_JUMP:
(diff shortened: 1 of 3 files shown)


recirq/quantum_chess/quantum_board.py
```diff
@@ -808,8 +809,6 @@ def do_move(self, m: move.Move) -> int:
             tqubit2 = bit_to_qubit(tbit2)
             self.add_entangled(squbit, tqubit, tqubit2)
             self.circuit.append(qm.split_move(squbit, tqubit, tqubit2))
-            self.state = set_nth_bit(sbit, self.state, False)
-            self.unhook(squbit)
```
dstrain115, line 812: Maybe we could change this to only unhook if the target squares were unentangled before moving?

Otherwise, this will leave a lot of entangled squares on the board which will be a source of unmitigated noise.

losos0: OK, I added the optimization to unhook in the unentangled 100 state.

In general there seems to be a lot of low-hanging fruit for this sort of qubit-saving operations (e.g. a path qubit is not needed for a split slide if there are no obstructing qubits on one path).

dstrain115: We should consider making a bug for these sorts of optimizations.  Seems like something a 20%er could easily tackle.

dstrain115 (approved):

PR #205, opened 2021-08-29
More tweaks for flaky tests

losos0:
In particular, use a fixed seed for the CI. I would have done this before, but I didn't realize the tests are actually running there.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -51,7 +51,8 @@ jobs:
     - name: Test with pytest
       run: |
         pip install pytest
-        pytest -v
+        # RECIRQ_CHESS_TEST_SEED: make quantum_chess tests determnistic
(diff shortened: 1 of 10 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install pytest
-        pytest -v
+        RECIRQ_CHESS_TEST_SEED=789 pytest -v
```
dstrain115, line 54: Maybe add a comment here to explain what this is for, in case people come here for a different part of reCirq.

dstrain115 (approved):

.github/workflows/pythonpackage.yml
losos0: Done.

PR #206, opened 2021-08-31
Make quantum board more automation-friendly

losos0:
Some tweaks I need for the fidelity testing script.

- Avoid slow ascii art rendering
- Let the user fully control the number of samples when needed

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -167,12 +167,34 @@ class CirqBoard:
         self.debug_log += (f"{action} takes {t1 - t0:0.4f} seconds.\n")
         self.timing_stats[action].append(t1 - t0)
 
-    def sample_with_ancilla(self, num_samples: int
-                            ) -> Tuple[List[int], List[Dict[str, int]]]:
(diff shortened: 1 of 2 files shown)


recirq/quantum_chess/quantum_board.py
```diff
@@ -167,19 +167,40 @@ def record_time(self, action: str, t0: float, t1: Optional[float] = None) -> Non
...
+            sample_size *= 2 ** (len(self.post_selection) + 1)
+        if self.error_mitigation == enums.ErrorMitigation.Correct:
+            sample_size *= 2
+        if self.noise_mitigation > 0:
+            sample_size *= 3
+        if sample_size < 100:
+            sample_size = 100
+        return sample_size
+
+    def sample_with_ancilla(self, num_samples: int, num_reps: int = -1
```
dstrain115, line 187: Why default of "-1"?  Why not set to Optional[int] = None, which would be more clear?

dstrain115 (approved):

recirq/quantum_chess/quantum_board.py
losos0: OK I changed it.

PR #207, opened 2021-08-31
Quantum chess bug fixes

losos0:
1. Fix equality test for `Move`
2. In the interactive board, don't log measurement statistics about the initial position
3. Fix a bug where a qubit with only single-qubit gates can be mapped to multiple grid qubits

### DIFF
--- a/recirq/quantum_chess/circuit_transformer.py
+++ b/recirq/quantum_chess/circuit_transformer.py
@@ -264,26 +264,22 @@ class ConnectivityHeuristicCircuitTransformer(CircuitTransformer):
         # Two qubit gates will turn into edges in the graph
         g = {}
         # Keep track of single qubits that don't interact.
-        sq = []
+        sq = set()
         for m in circuit:
             for op in m:
                 if len(op.qubits) == 1:
-                    if op.qubits[0] not in g:
-                        sq.append(op.qubits[0])
-                if len(op.qubits) == 2:
+                    sq.add(op.qubits[0])
+                else:
                     q1, q2 = op.qubits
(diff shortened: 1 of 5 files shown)


dstrain115 (commented):
Can you add a description to the PR to explain the bugs?  I'm not totally sure I understand what the problem was.

Also, I don't see any regression tests that test them.  The usual practice is to put a failing test that is fixed by the change.

losos0:
Added description to the OP.

- (1) already has a test.
- (2) is about logging debugging information so there is no test. I added additional description to the commit message to clarify.
- Added a test for (3).

recirq/quantum_chess/circuit_transformer.py
```diff
@@ -323,8 +319,7 @@ def qubit_mapping(self,
             mapping[cur_node] = start_qubit
             start_list.remove(start_qubit)
 
-        if len(mapping) != len(g):
-            print('Warning: could not map all qubits!')
+        assert len(mapping) == len(g) + len(sq)
```
dstrain115, line 322: Should we assert with a message so that people get a message?
Alternatively, should this raise an appropriate Exception instead?

dstrain115 (approved):
LGTM after a minor comment.

recirq/quantum_chess/circuit_transformer.py
losos0: There is always supposed to be a one-to-one mapping between named qubits and allocated grid qubits. An assertion seems appropriate to me since if it does not hold it means there is a bug in qubit_mapping.

Added a message to the assertion

PR #215, opened 2021-09-20
Prepare for Cirq 0.12

### LINKED ISSUE #200: Update to Cirq 0.12
A new Cirq has been released. I will track any outstanding migration issues here.

mpharrigan:
Towards #200.

This is blocked by openfermion and/or pytket releasing versions that support Cirq 0.12, although if you manually install Cirq 0.12 with these changes and an older pip that is ok with conflicting version specs, everything seems to work.

- Most changes are fixing deprecation warnings
- The one fix where recirq was fully broken was changing the `cirq_google.gate_sets` import
- Manually tested notebooks, including pre-filled-out ones


mpharrigan:
I don't know why the github actions are hanging with no output at random points in the process

mpharrigan:
I can reproduce this locally with pip 21.2.4 instead of pip 20.0.2 which just emits a warning

```
ERROR: pytket 0.13.0 has requirement jinja2~=2.11, but you'll have jinja2 3.0.1 which is incompatible.
ERROR: cirq 0.11.1 has requirement cirq-core==0.11.1, but you'll have cirq-core 0.12.0 which is incompatible.
ERROR: cirq 0.11.1 has requirement cirq-google==0.11.1, but you'll have cirq-google 0.12.0 which is incompatible.
```

mpharrigan:
I have to comment both the pytket and openfermion dependencies, otherwise it gets in a loop about `babel` and `tqdm`. I can't get it into a situation where it says a cirq version conflict is causing it

1) pip resolver is very bad and can't report on this as the cause of the conflict
2) there's some weird jinja mismatch between cirq, openfermion, and pytket that I don't have the complete picture on

mpharrigan:
@dstrain115 @MichaelBroughton PTAL. This doesn't actually bump the requirement but changes the few things that are necessary to have it work with the latest Cirq if you override pip.

dstrain115 (approved):
Changes look fine to me.

Issue #213, opened 2021-09-20
quantum_board_test is slow


mpharrigan:
This test module takes up 50% of total recirq unit test time on my computer. It's not excessive at this point, but I thought it would be worth asking if we could get the same coverage with less time.

@dstrain115 @cantwellc

dstrain115:
Some of the bigger tests involve larger number of qubits (10-12 qubits or so), so we could potentially remove noisy simulations for those tests and/or try to reduce them even further.

PR #219, opened 2021-09-26
Remove some slow-running tests

dstrain115:
- mpharrigan noticed that quantum_board_test was taking most of the
time for unit testing out of the whole recirq package.
- This PR remvoes a few of the slowest tests in order to speed up the
whole pytest run.
- Most of the removed tests are just removing some of the devices for
the bigger circuits.  Since qubit placement for devicesshould be tested in
multiple places, I don't really think we are losing much coverage by
removing these tests.
- On my system, this improves testing speed of this file from 56 seconds
to about 25 seconds.

Fixes: #213

### DIFF
--- a/recirq/quantum_chess/quantum_board_test.py
+++ b/recirq/quantum_chess/quantum_board_test.py
@@ -138,8 +138,8 @@ def test_path_qubits():
         b.path_qubits("c4", "a1")
 
 @pytest.mark.parametrize('move_type,board', (
-        *[(enums.MoveType.SPLIT_JUMP, b) for b in ALL_CIRQ_BOARDS],
-        *[(enums.MoveType.SPLIT_SLIDE, b) for b in ALL_CIRQ_BOARDS],
+        *[(enums.MoveType.SPLIT_JUMP, b) for b in BIG_CIRQ_BOARDS],
+        *[(enums.MoveType.SPLIT_SLIDE, b) for b in BIG_CIRQ_BOARDS],
 ))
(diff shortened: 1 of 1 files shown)


weinstein (approved):
Do we need to test on the small boards at all? As you say, circuit placement _should_ be tested separately. And future maintainers adding new tests will probably not know how to pick between the big boards, all boards, or just the simulator. Can we make that decision easier by eliminating some choices and/or documentation?

mpharrigan:
Thanks for doing this @dstrain115 ! 

I'm going to merge, but please continue the discussion @weinstein started

losos0:
The 4 testing boards we have vary on a few dimensions:
- Whether the gates in the quantum circuit are used as written, or converted to the Sycamore gate set
- Which, if any, device layout is used when placing qubits
- Whether noise (plus noise/error mitigation) is used

| Board Simulator | Gates | Layout | Noise | 
| --- | --- | --- | --- |
| `simulator` | As written | None | No |
| `syc54_noiseless` | Sycamore | Syc54 | No |
| `syc23_noiseless` | Sycamore | Syc23 | No |
| `syc23_noisy` | Sycamore | Syc23 | Yes |

A small board is needed to test layout; there are no other tests for layout regressions using real quantum chess circuits. However, `syc23_noiseless` seems somewhat redundant. If we got rid of `syc23_noiseless` then it seems like we wouldn't lose much coverage, assuming `simulator` and `syc23_noisy` are running.

PR #222, opened 2021-10-19
Intermediate starting states

MattKaplan:
with a few caching tests that fail


losos0:
The caching tests fail because `is_classical` calls `get_probability_distribution` with sample size 1000 and caching disabled. The probabilities are stored in `self.move_history_probabilities_cache`. Then when `get_probability_distribution` is called the next time (with caching enabled), it sees that there are already probabilities available with a large enough sample size and does not recalculate.

To avoid such inconsistent results, I think `use_cache` should be changed to a property of the QuantumBoard object so that caching in consistently enabled or disabled, instead of an argument that can be plumbed through a bunch of different functions and can have different values at any time.

losos0:
Unrelatedly, the `helper_board` object in cache_results should be initialized with `reset_starting_states=False` so that in case the default value is changed to True, the secondary board doesn't do any unwanted samples.

losos0:
In #226 I'm removing the use_cache arguments (so it's always on). But I believe you'll need some slight changes to the caching test for it to pass with 'reset starting states' enabled. The caching test uses a sample size of 100, and is_classical uses a sample size of 1000, which means the sample taken in is_classical will be used since it is bigger. So there's 2 ways it could be fixed:
- Scenario 1: change caching sample sizes to something larger than 1000. This will force a new sample to be taken after the is_classical one.
- Scenario 2: use sample size of exactly 1000 in the test. Then the sample taken by is_classical will use caching and no more samples will be needed after do_move is called.

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -1412,3 +1412,71 @@ def test_split_capture_with_failed_measurement_outcome(board):
...
+    assert b.is_classical()
+    assert b.move_history != []
+    assert b.circuit == cirq.Circuit()
+    assert b.ancilla_count == 0
+
+
+@pytest.mark.parametrize("board", ALL_CIRQ_BOARDS)
+def test_measurement_without_fully_classical_position(board):
+    """Splits piece on d3 to d7 and h3, then splits piece on b5 to b7 and b3.
+    Piece on f5 then moves to c8. Piece on c8 then moves to h3, with failed measurement."""
```
losos0, line 1473: > Piece on c8 then moves to h3, with failed measurement

All moves here have the BASIC move variant, which means no measurement is performed (unless the move type is castling). CAPTURE  or EXCLUDED variants do measurements

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -1412,3 +1412,71 @@ def test_split_capture_with_failed_measurement_outcome(board):
...
+def test_merge_to_fully_classical_position(board):
+    """Splits piece on d4 to b3 and c2, then merge back to fully classical position on a1."""
+    b = simulator(u.squares_to_bitboard(["d4"]))
+    b.reset_starting_states = True
+    assert b.perform_moves(
+        "d4^b3c2:SPLIT_JUMP:BASIC",
+        "b3c2^a1:MERGE_JUMP:BASIC",
+    )
+
+    assert b.is_classical()
```
losos0, line 1427: It would be good to check that the bitboard equals the expected one here

recirq/quantum_chess/quantum_board.py
```diff
@@ -431,13 +442,17 @@ def _caching_supported(self, m: move.Move):
     def cache_results(self, cache_key: CacheKey):
         if cache_key in self.cache:
             return
+
+        self.reset_starting_states = False
```
losos0, line 446: This shouldn't be here

recirq/quantum_chess/quantum_board.py
```diff
@@ -431,13 +442,17 @@ def _caching_supported(self, m: move.Move):
...
+
         if cache_key.move_type == enums.MoveType.SPLIT_JUMP:
             helper_board = CirqBoard(
                 DEFAULT_CHESS_INIT_STATE,
                 self.sampler,
                 self.device,
                 self.error_mitigation,
                 self.noise_mitigation,
+                self.reset_starting_states,
```
losos0, line 455: reset_starting_states=false?

MattKaplan: I tried doing it that way before, but it gave me  "SyntaxError: expression cannot contain assignment, perhaps you meant "=="." It works the way I have it, but happy to change it too, just not sure why it's giving me that error

recirq/quantum_chess/quantum_board_test.py
MattKaplan: I changed it so that it checks to make sure the probabilities are the same now

recirq/quantum_chess/quantum_board.py
losos0: Maybe you wrote `self.reset_starting_states=False`

MattKaplan: I tried both, and with "reset_starting_states=False", there's no error, but the caching tests still fail

losos0: Yeah, as I wrote in an earlier comment I don't expect the tests to pass without rebasing on #226.

MattKaplan: ah, makes sense. the PR is updated now

losos0:
OK, now #226 was merged.

MattKaplan:
>

recirq/quantum_chess/quantum_board.py
MattKaplan: mentioned this below but wanted to make sure you see - those few precache tests still fail after I committed the changes. Will look into it more

MattKaplan: Alright, should be good now

recirq/quantum_chess/quantum_board.py
```diff
@@ -119,19 +126,19 @@ def with_state(self, basis_state: int) -> "CirqBoard":
         self.post_selection = {}
         self.circuit = cirq.Circuit()
         self.ancilla_count = 0
-        self.move_history = []
         initial_square_probs = (float(nth_bit_of(i, basis_state)) for i in range(64))
-        # Invariant: len(self.move_history_probabilities_cache) ==
```
losos0, line 124: Don't delete the comment

recirq/quantum_chess/quantum_board.py
```diff
@@ -93,14 +94,15 @@ def __init__(
         ] = enums.ErrorMitigation.Nothing,
         noise_mitigation: Optional[float] = 0.0,
         transformer: Optional[ct.CircuitTransformer] = None,
+        reset_starting_states=True,
```
losos0, line 97: Don't forget to change the default to False before committing

losos0 (approved):

recirq/quantum_chess/quantum_board.py
```diff
@@ -37,6 +37,7 @@
 import recirq.quantum_chess.move as move
 import recirq.quantum_chess.quantum_moves as qm
 
+CLASSICAL_BITBOARD = 2 ** 64 - 1
```
dstrain115, line 40: Add a comment explaining what this is.

recirq/quantum_chess/quantum_board.py
```diff
@@ -93,14 +94,15 @@ def __init__(
...
+        reset_starting_states=True,
     ):
         self.device = device
         self.sampler = sampler
         if device is not None:
             self.transformer = (
                 transformer or ct.ConnectivityHeuristicCircuitTransformer(device)
             )
-        self.with_state(init_basis_state)
+        self.with_state(init_basis_state, True)
```
dstrain115, line 105: True is the default and is nit strictly needed.  Do you think keeping "True" makes things more clear?

recirq/quantum_chess/quantum_board.py
```diff
@@ -109,8 +111,13 @@ def __init__(
 
         self.cache = {}
 
-    def with_state(self, basis_state: int) -> "CirqBoard":
-        """Resets the board with a specific classical state."""
+        # Will only be turned on if user specifies
+        self.reset_starting_states = reset_starting_states
+
+    def with_state(self, basis_state: int, reset_move_history=True) -> "CirqBoard":
+        """Resets the board with a specific classical state. reset_move_history indicates
```
dstrain115, line 118: Probably Could be more clear to put the arguments into its own block:

```
"""Resets the board with a specific classical state.

Args:
    reset_move_history: indicates whether to reset the entire move history of the game. It will be set to false
        if we are calling this function after the board has returned to a fully classical position
"""
```

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -1412,3 +1414,76 @@ def test_split_capture_with_failed_measurement_outcome(board):
...
+    expected_probs = [0] * 64
+    expected_probs[square_to_bit("a1")] = 1
+    probs = list(b.get_probability_distribution(100))
+
+    assert b.is_classical()
+    assert probs == expected_probs
+
+
+@pytest.mark.parametrize("board", ALL_CIRQ_BOARDS)
+def test_undo_to_start_after_measurement(board):
```
dstrain115, line 1438: Maybe change to name to "test_undo_to_start_after_classical_reset"

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -1211,10 +1211,10 @@ def test_caching_accumulations_different_repetition_not_cached(board):
         )
     )
     probs1 = b.get_probability_distribution(1)
-    probs2 = b.get_probability_distribution(100)
+    probs2 = b.get_probability_distribution(1001)
```
dstrain115, line 1214: Optional:  it might make sense to change this into a constant so that it is clear that the number of reps is important.

For instance, CACHE_INVALIDATION_REPS maybe?  Then add a comment where you define the constant as to what this is.

recirq/quantum_chess/quantum_board.py
MattKaplan: Ah, I think that's left over from a time when true might not have been the default. It's probably not necessary to keep.

recirq/quantum_chess/quantum_board_test.py
```diff
@@ -33,6 +33,9 @@
 from recirq.quantum_chess.bit_utils import bit_to_qubit, square_to_bit, nth_bit_of
 from recirq.quantum_chess.caching_utils import CacheKey
 
+# The number of reps needed to avoid having the sample size of is_classical(), which is 1000,
+# chosen over the sampling in the caching tests
```
dstrain115, line 37: nit: This comment probably needs rephrasing.  It's confusing.  Maybe:

```
# The number of samples needed to avoid caching previous repetitions.
# Choosing this to be above the sample size of is_classical(), which is 1000, will avoid caching.
```

recirq/quantum_chess/quantum_board.py
```diff
@@ -109,8 +115,16 @@ def __init__(
...
-    def with_state(self, basis_state: int) -> "CirqBoard":
-        """Resets the board with a specific classical state."""
+        # Will only be turned on if user specifies
+        self.reset_starting_states = reset_starting_states
+
+    def with_state(self, basis_state: int, reset_move_history=True) -> "CirqBoard":
+        """Resets the board with a specific classical state.
+
+        Args:
+            reset_move_history: indicates whether to reset the entire move history of the game. It will be set to false
```
dstrain115, line 125: Also add basis_state.

dstrain115 (approved):
LGTM after a few minor nits.  Thanks for working so hard to get this over the finish line and contributing to quantum chess!

PR #226, opened 2021-11-05
Remove use_cache arguments

losos0:
We discussed making 'use cache' a class property, but actually the pre-cached move feature can't do anything anyway, unless you feed it with cached moves. So it seems unnecessary to a have a switch turning it off at all - just don't feed it with pre-cached moves. (See commit message for more details).

### DIFF
--- a/recirq/quantum_chess/caching_utils.py
+++ b/recirq/quantum_chess/caching_utils.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from typing import Tuple
 import recirq.quantum_chess.enums as enums
 import recirq.quantum_chess.move as move
 
@@ -11,3 +12,20 @@ class CacheKey:
 
 def cache_key_from_move(m: move.Move, repetitions: int) -> CacheKey:
     return CacheKey(m.move_type, repetitions)
+
+
+@dataclass(frozen=True)
+class ProbabilityHistory:
+    """Stores square occupancy histogram for a point in the move history.
+
+    Args:
+        repetitions: number of samples used to generate probabilities
+        probabilities: maps square -> probability of being occupied
+        full_squares: (derived from self.probabilities) full squares bitboard
+        empty_squares: (derived from self.probabilities) empty squares bitboard
+    """
+
+    repetitions: int
+    probabilities: Tuple[float, ...]  # for each square
+    full_squares: int  # bitboard derived from square_probabilities
+    empty_squares: int  # bitboard derived from square_probabilities
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -41,6 +41,13 @@ import recirq.quantum_chess.quantum_moves as qm
 # This is the basis state corresponding to FEN: rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR.
 DEFAULT_CHESS_INIT_STATE = 0xFFFF00000000FFFF
 
+# Number less than 0, so that the repetitions count of the cache entry is
+# always considered to be insufficient
+_NO_CACHE_AVAILABLE = -1
+
+# Repetitions count for position which should always use the cache
+_CACHE_ALWAYS_AVAILABLE = 10 ** 9
+
 
 class CirqBoard:
     """Implementation of Quantum Board API using cirq sampler.
@@ -98,13 +105,13 @@ class CirqBoard:
         self.noise_mitigation = noise_mitigation
 
         # Stores the repetition number if there is a cache.
-        self.board_accumulations_repetitions = -1
+        self.board_accumulations_repetitions = _NO_CACHE_AVAILABLE
 
         self.cache = {}
 
     def with_state(self, basis_state: int) -> "CirqBoard":
         """Resets the board with a specific classical state."""
-        self.board_accumulations_repetitions = -1
+        self.board_accumulations_repetitions = _NO_CACHE_AVAILABLE
         self.state = basis_state
         self.allowed_pieces = set()
         self.allowed_pieces.add(num_ones(self.state))
@@ -117,7 +124,9 @@ class CirqBoard:
         # Invariant: len(self.move_history_probabilities_cache) ==
         #            len(self.move_history) + 1
         self.move_history_probabilities_cache = [
-            self._make_probability_history(10 ** 9, initial_square_probs)
+            self._make_probability_history(
+                _CACHE_ALWAYS_AVAILABLE, initial_square_probs
+            )
         ]
 
         # Store the initial basis state so that we can use it for replaying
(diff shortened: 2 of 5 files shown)


recirq/quantum_chess/caching_utils.py
```diff
@@ -11,3 +12,11 @@ class CacheKey:
 
 def cache_key_from_move(m: move.Move, repetitions: int) -> CacheKey:
     return CacheKey(m.move_type, repetitions)
+
+
+@dataclass(frozen=True)
+class ProbabilityHistory:
```
dstrain115, line 18: Add doc string with Params rather than line comments.

recirq/quantum_chess/quantum_board.py
```diff
@@ -93,15 +97,14 @@ def __init__(
         self.error_mitigation = error_mitigation
         self.noise_mitigation = noise_mitigation
 
-        # None if there is no cache, stores the repetition number if there is a cache.
-        self.accumulations_repetitions = None
-        self.board_accumulations_repetitions = None
+        # Stores the repetition number if there is a cache.
+        self.board_accumulations_repetitions = -1
```
dstrain115, line 101: Why are we using -1 as a magic number to mean no repetitions rather than None?  I don't think this is a good idea.  Even if you persist in doing it, we should name it as a constant value.

recirq/quantum_chess/quantum_board.py
```diff
@@ -346,84 +345,75 @@ def sample(self, num_samples: int) -> List[int]:
...
+        if len(self.move_history) > 0:
             last_move = self.move_history[-1]
-            if self.move_history_probabilities_cache[-1][0] >= repetitions:
-                self.probabilities = self.move_history_probabilities_cache[-1][1].copy()
-                self._set_full_empty_squares_from_probability()
-                return
             previous_move_in_cache = (
-                len(self.move_history) > 1
-                and self.move_history_probabilities_cache[-2][0] >= repetitions
+                self.move_history_probabilities_cache[-2].repetitions >= repetitions
```
dstrain115, line 374: Won't this throw a KeyError if you get rid of the other conditions and it only has one entry?

recirq/quantum_chess/quantum_board.py
```diff
@@ -346,84 +345,75 @@ def sample(self, num_samples: int) -> List[int]:
...
-                    previous_probability = self.move_history_probabilities_cache[
-                        len(self.move_history) - 2
-                    ][1]
-                else:
-                    previous_probability = [0] * 64
-                    for i in range(64):
-                        # Assume initial state is classical
-                        previous_probability[i] = nth_bit_of(i, self.state)
+                previous_probability = self.move_history_probabilities_cache[
+                    -2
```
dstrain115, line 383: nit: this is some weird indenting.
Also, I am worried this may have a KeyError as well.

recirq/quantum_chess/quantum_board.py
losos0: After this refactoring, there is a cache entry for the initial position. So move_history_probabilities_cache is always longer by 1 element than move_history.

recirq/quantum_chess/quantum_board.py
losos0: Well you guys added the formatter check thing that enforces ugly formatting 😛 (even adding parentheses doesn't work, it deletes them)

recirq/quantum_chess/caching_utils.py
losos0: I'm not familiar with a Params section, but I put the stuff in a doc string.

recirq/quantum_chess/quantum_board.py
dstrain115: Optional: Is it worth documenting this behavior somewhere (maybe where we initialize move_history_probabilities_cache) so that it is clear what this structure is and how it works?

recirq/quantum_chess/caching_utils.py
dstrain115: I think the docstring below is supposed to be Args, not Attributes and have colons, not dashes:

```
Dataclass for storing histograms of board position probabilities from a given point in the move history.

Args:
    repetitions: number of samples used to generate probabilities
    probabilities: maps square -> probability of being occupied
    full_squares: (derived from self.probabilities) full squares bitboard
    empty_squares: (derived from self.probabilities) empty squares bitboard
```

recirq/quantum_chess/quantum_board.py
losos0: Because it is an int which is smaller than any possible number of repetitions. So it gives the right result when comparing the cached repetitions to the requested number of repetitions

recirq/quantum_chess/quantum_board.py
losos0: I did add a comment about this invariant in `__init__`

recirq/quantum_chess/caching_utils.py
losos0: Done.

recirq/quantum_chess/quantum_board.py
dstrain115: I would still make this a constant, such as UNINITIALIZED_CACHE_VALUE or something like that.

recirq/quantum_chess/caching_utils.py
dstrain115: Indentation is still not correct, and you are using dashes instead of colons.

recirq/quantum_chess/quantum_board.py
```diff
@@ -346,84 +347,75 @@ def sample(self, num_samples: int) -> List[int]:
             rtn = rtn + samples
         return rtn[:num_samples]
 
-    def _set_full_empty_squares_from_probability(self) -> None:
-        self.full_squares = 0
-        self.empty_squares = 0
-        for i, p in enumerate(self.probabilities):
+    def _make_probability_history(self, repetitions, probabilities):
```
dstrain115, line 350: missing type information.

losos0: Added

recirq/quantum_chess/quantum_board.py
losos0: Done

recirq/quantum_chess/caching_utils.py
losos0: All right. Is there some style guide for this repo that I should be consulting?

dstrain115: Here's the Google python style guide:
https://google.github.io/styleguide/pyguide.html#doc-function-args

The docstring formatting is so that any docs can be parsed and show up correctly on the documentation, though I don't think the reCirq stuff is currently listed on quantumai.google.

recirq/quantum_chess/quantum_board.py
```diff
@@ -110,14 +117,12 @@ def with_state(self, basis_state: int) -> "CirqBoard":
...
-            self.empty_squares = set_nth_bit(
-                i, self.empty_squares, not nth_bit_of(i, self.full_squares)
-            )
-        # Each entry is a 2-tuple of (repetitions, probabilities) corresponding to the probabilities after each move.
-        self.move_history_probabilities_cache = []
+        initial_square_probs = (float(nth_bit_of(i, basis_state)) for i in range(64))
+        # Invariant:
+        # self.move_history_probabilities_cache == len(self.move_history) + 1
+        self.move_history_probabilities_cache = [
+            self._make_probability_history(10 ** 9, initial_square_probs)
```
dstrain115, line 124: 10**9 should also be a constant I think.  I doubt people will request more than 1 billion reps of the initial state, but it would still be good to list that as a constant at the top.

dstrain115 (approved):

PR #228, opened 2021-11-10
Add function for 2-qubit tomography

losos0:
two_square_density_matrix() gives the density matrix for any pair of
squares.

This requires running 9 variants of the circuit with different gates at
the end tweaking the two qubits of interest.

Error correction and noise mitigation are disabled, since the qubits
undergoing tomography can't be measured in the computational basis.

The density matrix can be viewed from interactive_board.py with the
command, e.g., `density f3 f5`.

---

Before merging I'd want to remove some of the duplicated code between the new function and `sample_with_ancilla`.


mpharrigan:
closing; should be re-routed to unitary/chess repository

PR #230, opened 2021-11-12
Move requirements into subpackages

mpharrigan:
- `extra-requirements.txt` per subpackage
 - Right now: union all the extra requirements into `install_requires` and add `extra_requires` for each subpackage (setup.py)
 - Next: modify notebooks to install using e.g. `recirq[qaoa]`. Remove extra requires from `install_requires`. This stepwise change is necessary since the notebooks install from github #231
 - Next: independent test jobs for subpackages to make sure the requirements are correct.

Annotated change:
```
# optimize
scikit-learn  # (moved to extra-requires)

# qaoa
networkx   # (transitive from cirq-core)
pytket-cirq~=0.16.0   # (moved to extra-requires)

# quantum chess, only needed for tests
scipy    # (transitive from cirq-core, test requirements shouldn't be here)

# hfvqe
openfermion~=1.2.0    # (moved to extra-requires)
# fix bug with openfermionpyscf https://github.com/quantumlib/ReCirq/issues/200#issuecomment-923203883
h5py~=3.2.1   # (moved to extra-requires)

# fermi_hubbard
tqdm  # notebook progress bars (transitive from cirq-core)

# otoc
Py-BOBYQA   # (moved to extra-requires)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ extras_require = {
     for r in extras_require
 }
 
-# TODO: remove and require users to install via extras_require.
+# TODO(gh-231): remove and require users to install via extras_require.
 install_requires = functools.reduce(operator.add, extras_require.values(), install_requires)
 
 setup(name='recirq',
--- a/setup.py
+++ b/setup.py
@@ -13,22 +13,41 @@
 # limitations under the License.
 
 import pathlib
+import functools
+import operator
+
 from setuptools import find_packages, setup
 
 __version__ = ''
 exec(open('recirq/_version.py').read())
 assert __version__, 'Version string cannot be empty'
 
-required_packages = pathlib.Path('requirements.txt').read_text().split('\n')
-INSTALL_PACKAGES = [pkg for pkg in required_packages if pkg and not pkg.startswith('#')]
+
+def _parse_requirements(path: pathlib.Path):
+    lines = [line.strip() for line in path.read_text().splitlines() if line]
+    return [line for line in lines if not line.startswith('#')]
+
+
+install_requires = _parse_requirements(pathlib.Path('requirements.txt'))
+extras_require = [
+    'otoc', 'qaoa', 'optimize', 'hfvqe', 'fermi_hubbard'
+]
+extras_require = {
+    r: _parse_requirements(pathlib.Path(f'recirq/{r}/extra-requirements.txt'))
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -13,22 +13,41 @@
...
+install_requires = _parse_requirements(pathlib.Path('requirements.txt'))
+extras_require = [
+    'otoc', 'qaoa', 'optimize', 'hfvqe', 'fermi_hubbard'
+]
+extras_require = {
+    r: _parse_requirements(pathlib.Path(f'recirq/{r}/extra-requirements.txt'))
+    for r in extras_require
+}
+
+# TODO: remove and require users to install via extras_require.
```
dstrain115, line 40: Should this link to a Github issue?

setup.py
```diff
@@ -13,22 +13,41 @@
...
-INSTALL_PACKAGES = [pkg for pkg in required_packages if pkg and not pkg.startswith('#')]
+
+def _parse_requirements(path: pathlib.Path):
+    lines = [line.strip() for line in path.read_text().splitlines() if line]
+    return [line for line in lines if not line.startswith('#')]
+
+
+install_requires = _parse_requirements(pathlib.Path('requirements.txt'))
+extras_require = [
+    'otoc', 'qaoa', 'optimize', 'hfvqe', 'fermi_hubbard'
```
dstrain115, line 33: Is there a good way to auto-detect this (or alternatively, for keeping this up to date if we add more experiments)?

dstrain115 (approved):
Seems fine but for two comments.

setup.py
mpharrigan: Yes, I first wrote it to glob/regex for extra-requirements.txt and I could set it to that. However: for my next thing I want to add CI jobs that tests each subpackage in isolation. Autogenerating CI jobs would likely be more complicated than its worth. Especially since there will be an isolated "vanilla deps" build that will run all tests _not_ in the named extra-requires subpackages which would fail if they did indeed have extra requirements.

setup.py
mpharrigan: done, also see #231

setup.py
dstrain115: Ok, SGTM, I already approved this PR.  Feel free to merge when ready.

PR #236, opened 2021-12-04
loschmidt.tilted_square_lattice Benchmark

mpharrigan:
Adds the executable generation functionality for the following benchmark:

```python
AlgorithmicBenchmark(
    domain='recirq.otoc',
    name='loschmidt.tilted_square_lattice',
    executable_family='recirq.otoc.loschmidt.tilted_square_lattice',
    spec_class=TiltedSquareLatticeLoschmidtSpec,
    executable_generator_func=get_all_tilted_square_lattice_executables,
)
```

The loschmidt echo runs a random unitary forward and backwards and measures how often we return to the starting state. This benchmark checks random unitaries generated on the TiltedSquareLattice topology.

### Notes

This PR does not include any BenchmarkConfig's, namely scripts for launching execution and analysis.

I've also removed the `data_class` field from the AlgorithmicBenchmark card catalog. (1) I don't want to make this PR too big by including it (2) I want to work some of the kinks out of the design of the data analysis part of the flow without holding up committing the infrastructure for taking data. 

## Cirq requirements

Following #237 , this will only test the new functionality against the pre-release version of cirq where the `cirq_google.workflow` utilities are available.

I've noticed as well that Cirq 0.12 does **not** include `cirq.NamedTopology`. This is a case where keeping compatibility with the prior version of Cirq is **not** worth it, esp. since v0.13 has been out for a while. As part of this PR I've disabled the "pre" check with a note. We'll re-enable when Cirq 0.14 comes out.

### DIFF
--- a/recirq/otoc/loschmidt/__init__.py
+++ b/recirq/otoc/loschmidt/__init__.py
@@ -1,0 +1,34 @@
+# Copyright 2021 Google
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from functools import lru_cache
+from typing import Optional
+
+from cirq.protocols.json_serialization import ObjectFactory, DEFAULT_RESOLVERS
+from .tilted_square_lattice import (
+    TiltedSquareLatticeLoschmidtSpec,
+)
+
+
+@lru_cache()
+def _resolve_json(cirq_type: str) -> Optional[ObjectFactory]:
+    if not cirq_type.startswith('recirq.otoc.'):
+        return None
+
+    cirq_type = cirq_type[len('recirq.otoc.'):]
+    return {k.__name__: k for k in [
+        TiltedSquareLatticeLoschmidtSpec,
+    ]}.get(cirq_type, None)
+
+
+DEFAULT_RESOLVERS.append(_resolve_json)
--- a/recirq/otoc/loschmidt/__init__.py
+++ b/recirq/otoc/loschmidt/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Google
+# Copyright 2022 Google
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice.py
+++ b/recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice.py
@@ -1,0 +1,146 @@
+import itertools
+from dataclasses import dataclass
+
+import numpy as np
+
+import cirq
+from cirq import TiltedSquareLattice
+from cirq.experiments import random_rotations_between_grid_interaction_layers_circuit
+from cirq.protocols import dataclass_json_dict
+from cirq_google.workflow import QuantumExecutable, BitstringsMeasurement, QuantumExecutableGroup, \
+    ExecutableSpec
+
+
+def create_tilted_square_lattice_loschmidt_echo_circuit(
+        topology: TiltedSquareLattice,
+        macrocycle_depth: int,
+        twoq_gate: cirq.Gate = cirq.FSimGate(np.pi / 4, 0.0),
+        rs: cirq.RANDOM_STATE_OR_SEED_LIKE = None,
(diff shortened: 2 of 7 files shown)


recirq/otoc/loschmidt/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import Optional
+
+from cirq.protocols.json_serialization import ObjectFactory, DEFAULT_RESOLVERS
+from .tilted_square_lattice import (
+    TiltedSquareLatticeLoschmidtSpec,
+)
+
+
+@lru_cache()
+def _resolve_json(cirq_type: str) -> Optional[ObjectFactory]:
```
dstrain115, line 24: I feel like I missed a change somewhere, since this is likely not the first time this snippet has appeared.
However, what is this?  I am confused why we need this.  Maybe a code comment would be good to explain what this is for.

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice.py
```diff
@@ -0,0 +1,147 @@
+import itertools
```
dstrain115, line 1: Missing copyright.

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
```
dstrain115, line 1: missing copyright.

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def mock_get_circuit(topology: cirq.TiltedSquareLattice, macrocycle_depth: int,
+                         twoq_gate: cirq.Gate = cirq.FSimGate(np.pi / 4, 0.0),
+                         rs: cirq.RANDOM_STATE_OR_SEED_LIKE = None):
+        nonlocal call_count
+        call_count += 1
+        return cirq.Circuit()
+
+    monkeypatch.setattr(recirq.otoc.loschmidt.tilted_square_lattice.tilted_square_lattice,
+                        "create_tilted_square_lattice_loschmidt_echo_circuit", mock_get_circuit)
```
dstrain115, line 60: Why do we need this monkeypatch?  Is it just for test performance?
If so, how slow is the test without this?

recirq/otoc/loschmidt/__init__.py
mpharrigan: this is how we register modules for json deserialization; done after the separation of `cirq_google` and friends. You can see we add it to cirq's `DEFAULT_RESOLVERS` global variable

mpharrigan: added a docstring

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
mpharrigan: done thanks

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
mpharrigan: It's for speed and so I can assert something, namely that it was called the correct number of times. 

Creating random circuits is expensive -- or at least speeding that up is outside the scope for this work. I think I wanted a test that relied on the default arguments for this function. I can add another test that dials back the number of executables and makes other assertions.

mpharrigan:
@dstrain115 I think I addressed your comments. PTAL if you have time

recirq/otoc/loschmidt/__init__.py
```diff
@@ -0,0 +1,39 @@
+# Copyright 2021 Google
```
dstrain115, line 1: Ha, it's 2022 now!  Gotcha!  
(optional since you started it in  2021)

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
dstrain115: Maybe add a comment to that effect above and explain what the test is actually testing.

dstrain115 (approved):
Minor comments then I approve.

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice_test.py
mpharrigan: added

recirq/otoc/loschmidt/__init__.py
mpharrigan: oh no! now people can skirt our onerous apache2 license because the copyright is invalid!

mpharrigan:
made the requested changes!

dstrain115 (approved):
LGTM

PR #237, opened 2021-12-04
CI Testing Matrix

### LINKED ISSUE #224: Update to Cirq 0.13
I'll track the effort needed to support Cirq 0.13 here.

 - [x] Lots of broken quantum chess tests: `circuit_transformer.py:286` fails when it encounters 3-qubit gates. I think `quantum_board.py:247` `SycamoreDecomposer` is not decomposing the tofflis @dstrain115
 - [x] Need openfermion that supports 0.13 https://github.com/quantumlib/OpenFermion/issues/751
 - [x] Verify latest pytket isn't over-pinned
 - [x] https://github.com/CQCL/pytket-extensions/issues/230 try to get pytket-cirq off of the cirq metapackage.
 - [x] Manually check notebooks

Other than quantum chess toffli check, all the tests pass.

mpharrigan:
## Problem overview

We depend on openfermion, pytket, and pytket-cirq; which have historically been over-pinned to a particular cirq version. "Overpinned" means that even if it seems to work with a newer (or older!) cirq version, the `install_requires` still specifies one version. `pip` has recently (past year) learned to be strict about satisfying requirements. `pip` will end up in an infinite loop downloading old versions instead of saying which package is causing a version deadlock

As a developer, I can get things to work by installing dependencies myself and then `pip install --no-deps` openfermion, cirq, pytket, or recirq so pip can't get in my way.

However, one important use case is supporting colab which must have a simple install script; hopefully just `pip install recirq` (or, specifically, the git url).

## Openfermion relaxation

Cirq has been generally stable across the past couple of releases, especially if one were to go through and fix deprecation warnings. Indeed, I have changed openfermion https://github.com/quantumlib/OpenFermion/pull/755/ to not pin in `install_requires` but to use the CI to test against multiple versions of Cirq.

## Towards relaxing ReCirq

This PR aims to head towards the same goal. We test against the previous, current, and next version of Cirq. Testing against the next version of Cirq is particularly important when committing code that depends on unreleased Cirq features. The trick is to _attempt_ the full matrix of versions, and lean heavily on `pytest.mark.skipif` to make sure tests only run under supported configurations. 

In particular, we introduce a script to capture the intricacies of the various test configurations: `write-ci-requirements.txt`. This lets us pin to a particular cirq version during CI testing but not in `install_requires`, which we change to a lower bound.

This script has several notable features
- translate fixed "relative" versions like current, previous, and next to numeric, pypi ones. Otherwise, the numeric versions from the CI matrix will show up in the job names. We want to set the "required checks" setting on GitHub without having it rely on the current cirq versions, which will change over time
- Hacky pytket logic: don't test the pytket functionality on the next cirq version.

## Cirq 0.13
This fixes #224 because pytket and openfermion have compatible versions now.

## Cirq 0.14
We'll now be in the regime where we need to fix ReCirq for the next Cirq version _before_ it is released. As such, this PR is based off of #241 .

## Per-package testing

This still tests the whole ReCirq repository at once. We can modify `write-ci-requirements.py` to help simplify a new matrix job that tests each subpackage in isolation. `skipif` may be a large component to that as well.

### DIFF
--- a/dev_tools/write-ci-requirements.py
+++ b/dev_tools/write-ci-requirements.py
@@ -1,0 +1,107 @@
+import pathlib
+import re
+import sys
+from argparse import ArgumentParser
+from typing import Any, Dict, List
+
+REPO_DIR = pathlib.Path(__file__).parent.parent.resolve()
+print('Repo dir:', REPO_DIR)
+
+CIRQ_VERSIONS = {
+    'previous': '==0.12.0',
+    'current': '==0.13.0',
+    'next': '>=0.14.0.dev',
+}
+"""Give names to relative Cirq versions so CI can have consistent names while versions 
+get incremented."""
+
+
+def _parse_requirements(path: pathlib.Path):
+    """Read and strip comments from a requirements.txt-like file. """
+    lines = [line.strip() for line in path.read_text().splitlines() if line]
+    return [line for line in lines if not line.startswith('#')]
+
+
+def _remove_version_spec(req: str) -> str:
+    """Remove a version specifier like ==1.3 from a requirements.txt line."""
+    components = re.split(r'>=|~=|<=|==', req)
+    return components[0]
+
+
+def _set_cirq_version(core_reqs: List[str], relative_cirq_version: str) -> List[str]:
+    """Return a new version of `core_reqs` that pins cirq-like packages to the desired version."""
+    cirq_version = CIRQ_VERSIONS[relative_cirq_version]
+    to_change = 'cirq', 'cirq-google', 'cirq-core'
+
+    new_reqs = []
+    for req in core_reqs:
+        without_spec = _remove_version_spec(req)
+        if without_spec in to_change:
+            new_reqs.append(f'{without_spec}{cirq_version}')
+        else:
+            new_reqs.append(req)
+
+    return new_reqs
+
+
+def _set_qaoa_hacks(qaoa_reqs: List[str], relative_cirq_version: str) -> List[str]:
+    """Pytket pins to a specific cirq version and doesn't work with cirq "next".
+    """
+    if relative_cirq_version != 'next':
+        return qaoa_reqs
+
+    new_reqs = []
+    for req in qaoa_reqs:
+        without_spec = _remove_version_spec(req)
+        if without_spec == 'pytket-cirq':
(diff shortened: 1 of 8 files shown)


mpharrigan:
rebased off of #241

dev_tools/write-ci-requirements.py
```diff
@@ -0,0 +1,107 @@
...
+from typing import Any, Dict, List
+
+REPO_DIR = pathlib.Path(__file__).parent.parent.resolve()
+print('Repo dir:', REPO_DIR)
+
+CIRQ_VERSIONS = {
+    'previous': '==0.12.0',
+    'current': '==0.13.0',
+    'next': '>=0.14.0.dev',
+}
```
dstrain115, line 14: Should we put a note in the requirements.txt file so that people know to go and update this?

recirq/qaoa/placement.py
```diff
@@ -3,17 +3,25 @@
...
-import networkx as nx
-import numpy as np
-import pytket
-import pytket.extensions.cirq
-from pytket.circuit import Node, Qubit
-from pytket.passes import SequencePass, RoutingPass, PlacementPass
-from pytket.predicates import CompilationUnit, ConnectivityPredicate
-from pytket.routing import GraphPlacement
+
+try:
```
dstrain115, line 13: Maybe a quick comment here to explain that this is here for CI tests?

dstrain115 (approved):
Cool!  One step towards a better ecosystem!

dev_tools/write-ci-requirements.py
mpharrigan: good idea, done

recirq/qaoa/placement.py
mpharrigan: added:
```
    # Set the 'RECIRQ_IGNORE_PYTKET' environment variable to treat PyTket as an optional
    # dependency. We do this for CI testing against the next, pre-release Cirq version.
```

(in general, users could also set this environment variable if they get themselves into hot water managing the pytket dependency)

mpharrigan:
I modified the `master` branch protection rules to use the new test names. I also added the notebook formatting test, which was never a required test

PR #238, opened 2021-12-04
Adds state learning experiment qml_lfe 4/n.

MichaelBroughton:
Adds the first of two state learning experiments to qml_lfe, with another appropriate fuzz test.

### DIFF
--- a/recirq/qml_lfe/learn_states_q.py
+++ b/recirq/qml_lfe/learn_states_q.py
@@ -1,0 +1,192 @@
+# Copyright 2021 Google
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Draw I + P pauli product data using two copy strategy.
+style circuits.
+
+learn_states_q.py --n=6 --n_paulis=20 --batch_size=500 --n_shots=1000
+
+Will create paulistring circuits all on 12 qubits (two systems of size 6).
+Once the circuits are generated, batch_size sweeps and a single circuit at
+a timewill be sent will be sent for simulation/execution, drawing n_shots
+total samples from each one using `run_sweep` in Cirq. By default the
+bitstring data will be saved in the data folder. One can also set
+`use_engine` to True in order to run this against a processor on
+quantum engine.
+"""
+from typing import Dict, List
+
+import os
+import cirq
+import numpy as np
+import sympy
+from . import circuit_blocks
+from . import run_config
+
+from absl import app
+from absl import logging
+
+
+def _create_randomized_sweeps(hidden_p: str, symbols: Dict[str, int], num_reps: int):
+    last_i = 0
+    for i, pauli in enumerate(hidden_p):
+        if pauli != "I":
+            last_i = i
+
+    sign_p = -1 if np.random.random() < 0.5 else 1
+    all_sweeps = []
+    for _ in range(num_reps):
+        current_sweep = dict()
+        for twocopy in [0, 1]:
+            parity = sign_p * (1 if np.random.random() <= 0.95 else -1)
+            for i, pauli in enumerate(hidden_p):
+                current_symbol = symbols[2 * i + twocopy]
+                current_sweep[current_symbol] = 0
+                if pauli != "I":
+                    if last_i == i:
+                        if parity == -1:
+                            current_sweep[current_symbol] = 1
+                    elif np.random.choice([0, 1]) == 1:
+                        parity *= -1
+                        current_sweep[current_symbol] = 1
+                else:
+                    if np.random.choice([0, 1]) == 1:
+                        current_sweep[current_symbol] = 1
+        all_sweeps.append(current_sweep)
+
+    return all_sweeps
+
+
+def build_circuit(qubit_pairs: List[List[cirq.Qid]], pauli: str, num_reps: int):
+    """Create I + P problem circuit between qubit pairs.
+
+    Args:
+        qubit_pairs: List of qubit pairs.
+        pauli: Python str containing characters 'I', 'X', 'Y' or 'Z'.
+        num_reps: Number of repetitions to generate for sweeps.
+
+    Returns:
+        A (circuit, sweep) tuple, runnable using `run_sweep`.
+    """
+    a_qubits = [pair[0] for pair in qubit_pairs]
+    b_qubits = [pair[1] for pair in qubit_pairs]
+    all_qubits = np.concatenate(qubit_pairs)
+
+    flip_params = sympy.symbols(f"param_0:{len(qubit_pairs) * 2}")
+
+    # Add X flips.
+    ret_circuit = cirq.Circuit(cirq.X(q) ** p for q, p in zip(all_qubits, flip_params))
+
+    # Add basis turns a and b.
+    ret_circuit += [
+        circuit_blocks.inv_z_basis_gate(p)(q) for q, p in zip(a_qubits, pauli)
+    ]
+    ret_circuit += [
+        circuit_blocks.inv_z_basis_gate(p)(q) for q, p in zip(b_qubits, pauli)
+    ]
+
+    # Add un-bell pair.
+    ret_circuit += [circuit_blocks.un_bell_pair_block(pair) for pair in qubit_pairs]
+
+    # Add measurements.
+    for i, qubit in enumerate(all_qubits):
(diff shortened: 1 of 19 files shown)


recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,194 @@
...
+from . import circuit_blocks
+from . import run_config
+
+from absl import app
+from absl import logging
+
+
+def _create_randomized_sweeps(
+    hidden_p: str, symbols: Dict[str, int], num_reps: int
+) -> List[Dict[str, int]]:
```
dstrain115, line 43: Even though this is a private function, it's fairly lengthy.  Might be worth a docstring to explain a bit, especially since it is not returning a sweep (as per the function name) but instead a list of dicts.

recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,194 @@
...
+    last_i = 0
+    for i, pauli in enumerate(hidden_p):
+        if pauli != "I":
```
dstrain115, line 46: Can you use `hidden_p.rfind("I")`?

recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,194 @@
...
+            last_i = i
+
+    sign_p = -1 if np.random.random() < 0.5 else 1
+    all_sweeps = []
+    for _ in range(num_reps):
+        current_sweep = dict()
+        for twocopy in [0, 1]:
+            parity = sign_p * (1 if np.random.random() <= 0.95 else -1)
```
dstrain115, line 54: What is the significance of 0.95 here?

recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,194 @@
...
+            for j, single_circuit_samples in enumerate(results):
+                qubit_order = [f"q{i}" for i in range(2 * n)]
+                out0 = single_circuit_samples.data[qubit_order].to_numpy()
+                batch_results.append(np.squeeze(out0))
+
+            batch_results = np.array(batch_results)
+            all_results.append(batch_results)
+
+        all_results = np.concatenate(all_results)
+        name0 = "Q-size-{}-pauli-{}".format(n, "".join(t for t in pauli))
```
dstrain115, line 175: nit:  why name0?  Would filename be a better name?

recirq/qml_lfe/learn_states_q_test.py
```diff
@@ -0,0 +1,58 @@
...
+# limitations under the License.
+
+import pytest
+import os
+import numpy as np
+from . import learn_states_q
+
+
+def _predict_exp(data, pauli_list, two_copy):
+    ave = 0
```
dstrain115, line 22: re: "ave"  Avoid abbreviations.  go/pystyle#naming

recirq/qml_lfe/learn_states_q_test.py
```diff
@@ -0,0 +1,58 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import os
+import numpy as np
+from . import learn_states_q
+
+
+def _predict_exp(data, pauli_list, two_copy):
```
dstrain115, line 21: This function is also hard to understand.  Maybe a docstring here to explain it might help?

recirq/qml_lfe/run_config.py
```diff
@@ -59,6 +61,30 @@ def execute_batch(
...
+        circuit: the `cirq.Circuit` with symbols to execute.
+        params: The list of parameters to place in the symbols.
+        use_engine: Whether or not to make use of quantum engine and the
+            weber processor. Note this requires the GOOGLE_CLOUD_PROJECT
+            environment variable to be set, along with the required cloud
+            permissions.
+    Returns:
+        List of results to execution.
+    """
+    runner = cirq.Simulator()
```
dstrain115, line 79: not: Maybe move this to an "else" block to avoid instantiation in the engine case?

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: Added.

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: One could do that, but to handle the case when there are no "I"'s means the logic needs to be osmething like:
```python3
last_i = hidden_p.rfind('I') - 1
if last_i < 0:
  last_i = len(hidden_p)
```
Which doesn't make it any shorter, so I left it as is.

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: This is a detail in the paper. We are actually generating rho = I +- 0.95 P (details in section A.2.a) https://arxiv.org/pdf/2112.00778.pdf

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: Changed.

recirq/qml_lfe/learn_states_q_test.py
MichaelBroughton: Good catch. Changed.

recirq/qml_lfe/learn_states_q_test.py
MichaelBroughton: Added a short docstring.

recirq/qml_lfe/run_config.py
MichaelBroughton: I had figured empty simulator instantiation was pretty light, but I've moved to an else block just to be safe.

recirq/qml_lfe/learn_states_q.py
mpharrigan: in general, comments referring to a specific section in the paper are highly encouraged

mpharrigan: ah, I see this is in the docstring.

mpharrigan: This, and above might read better if you 

1) allow optional passing of a numpy randomstate like in cirq for optional determinism
2) use `rs.choice([1,-1], p=[0.95, 0.05])`

mpharrigan: especially since you use `choice` later

recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,209 @@
...
+    for _ in range(num_reps):
+        current_sweep = dict()
+        for twocopy in [0, 1]:
+            parity = sign_p * (1 if np.random.random() <= 0.95 else -1)
+            for i, pauli in enumerate(hidden_p):
+                current_symbol = symbols[2 * i + twocopy]
+                current_sweep[current_symbol] = 0
+                if pauli != "I":
+                    if last_i == i:
+                        current_sweep[current_symbol] = 1 if parity == -1 else 0
```
mpharrigan, line 75: this code is _very_ highly nested. any way to factor out? or use `np.product` instead of nested for loops?

recirq/qml_lfe/run_config.py
```diff
@@ -59,6 +63,32 @@ def execute_batch(
...
+        use_engine: Whether or not to make use of quantum engine and the
+            weber processor. Note this requires the GOOGLE_CLOUD_PROJECT
+            environment variable to be set, along with the required cloud
+            permissions.
+    Returns:
+        List of results to execution.
+    """
+    runner = None
+    if use_engine:
+        runner = cirq_google.get_engine_sampler(
```
mpharrigan, line 83: why not pass around `runner` instead of a `use_engine` flag?

recirq/qml_lfe/learn_states_q.py
```diff
@@ -0,0 +1,209 @@
...
+# limitations under the License.
+
+"""Draw I + P pauli product data using two copy strategy.
+style circuits.
+
+learn_states_q.py --n=6 --n_paulis=20 --batch_size=500 --n_shots=1000
+
+Will create paulistring circuits all on 12 qubits (two systems of size 6).
+Once the circuits are generated, batch_size sweeps and a single circuit at
+a timewill be sent will be sent for simulation/execution, drawing n_shots
```
mpharrigan, line 22: timewill -> time will

mpharrigan (approved):
pre-emptive LGTM, would prefer if you reduced nesting in that one function

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: Changed this to make use of randomstate. I was a little lax with integrating this in some other places in the code so I may circle back and improve later.

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: Did my best to un-nest, still not great, but better than before.

recirq/qml_lfe/learn_states_q.py
MichaelBroughton: Fixed.

recirq/qml_lfe/run_config.py
MichaelBroughton: I guess I could do that. I'll look into it in a follow on PR.

PR #241, opened 2021-12-06
Changes for Cirq 0.14

mpharrigan:
- cirq_google
- EngineTimeSlot removal. If this helper function is necessary for quantum chess, please open a follow-on PR to re-enable cc @dstrain115 
- TwoQubitGate removal, used in the tests

### DIFF
--- a/recirq/engine_utils.py
+++ b/recirq/engine_utils.py
@@ -20,11 +20,11 @@ import uuid
 from dataclasses import dataclass, field
 from typing import List, Any, Optional, Callable, Dict, Union
 
+import numpy as np
+
 import cirq
 import cirq_google as cg
-import numpy as np
 from cirq import work, study, circuits, ops
-from cirq_google.engine.client.quantum_v1alpha1.gapic import enums
 from cirq_google.engine.engine_job import TERMINAL_STATES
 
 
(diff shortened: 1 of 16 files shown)


recirq/engine_utils.py
```diff
@@ -377,42 +377,3 @@ async def worker():
...
-    current_time = _get_current_time()
-    for processor_name in processor_names:
-        processor_id = get_processor_id_by_device_name(processor_name)
-        if processor_id is None:
-            # Skip the check if this is a simulator.
-            continue
-        processor = engine.get_processor(processor_id)
-        for time_slot in processor.get_schedule():
-            try:
-                time_slot = cg.EngineTimeSlot.from_proto(time_slot)
```
dstrain115, line 403: It looks like nothing is using this function, so I guess it's okay to delete it.  I think the goal was to have a quantum chess interface that could opportunistically use open swim slots to run moves when the processor was available, but I doubt that will happen now.

If you want to keep this function, you can just get rid of `cg.EngineTimeSlot.from_proto(time_slot)` and operate directly on the `QuantumTimeSlot` which should be very similar.  I think if you just delete this line, it might work (or be very close to working).  If you want to take this opportunity to prune the dead code though, that's also fine with me.

dstrain115 (approved):
LGTM with an optional comment if you want to try to salvage the function.

recirq/engine_utils.py
mpharrigan: I think `get_schedule` has the wrong return type annotation. why doesn't mypy pick it up?

mpharrigan: I got 

```
E               TypeError: '<' not supported between instances of 'Timestamp' and 'datetime.datetime'
```

I tried changing 
```
def _get_current_time():
    from google.protobuf.timestamp_pb2 import Timestamp
    timestamp = Timestamp()
    return timestamp.GetCurrentTime()
```

which confusingly did not change the error. I think I'm going to give up and continue with the removal.

PR #243, opened 2021-12-07
Fix failing tensorflow.org import job.

MarkDaoust:
This notebook uses cirq_google, but forgot to import it.


dstrain115 (approved):
Thanks for proactively fixing this!

PR #245, opened 2021-12-09
Added README.md file to qml_lfe 6/7.

MichaelBroughton:
Also removed some relative imports.

### DIFF
--- a/recirq/qml_lfe/__init__.py
+++ b/recirq/qml_lfe/__init__.py
@@ -1,0 +1,6 @@
+# Insert qml_lfe on PYTHONPATH. This is to avoid the requirement
+# of needing to do from recirq.qml_lfe.xxx style imports inside of the
+# qml_lfe directory.
+import sys
+from pathlib import Path
+sys.path.insert(0, str(Path(__file__).parent))
--- a/recirq/qml_lfe/__init__.py
+++ b/recirq/qml_lfe/__init__.py
@@ -1,6 +1,0 @@
-# Insert qml_lfe on PYTHONPATH. This is to avoid the requirement
-# of needing to do from recirq.qml_lfe.xxx style imports inside of the
-# qml_lfe directory.
-import sys
-from pathlib import Path
-sys.path.insert(0, str(Path(__file__).parent))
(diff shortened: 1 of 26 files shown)


recirq/qml_lfe/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Insert qml_lfe on PYTHONPATH. This is to avoid the requirement
+# of needing to do from recirq.qml_lfe.xxx style imports inside of the
+# qml_lfe directory.
+import sys
+from pathlib import Path
+sys.path.insert(0, str(Path(__file__).parent))
```
dstrain115, line 6: This seems suspicious.  I am not sure if the path trickery is worth it if it is just to make imports look nicer.

dstrain115 (commented):
Very excited to learn that n=7!

recirq/qml_lfe/__init__.py
MichaelBroughton: I'm not a huge fan either, but I do think it is our best option. In certain cases if I called this module from different directories (other than directly inside qml_lfe) things might break. I'm also not a huge fan of doing `from recirq.qml_lfe.whatever import xxx` because it makes the code less standalone and self contained, which makes things harder than they need to be for a user who might want to pull this project out and hack on it.

mpharrigan: veto on this

recirq/qml_lfe/circuit_blocks_test.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import pytest
-from . import circuit_blocks
+import circuit_blocks
```
mpharrigan, line 16: this isn't how we do imports in python

mpharrigan (changes requested):
please do normal import semantics

MichaelBroughton:
Reverted to `from recirq.qml_lfe.xxx import whatever` after offline discussion.

recirq/qml_lfe/__init__.py
MichaelBroughton: Alrighy, moved to using proper imports as pert style. It is a little akward launching off the modules now, but such is life.

mpharrigan (approved):

PR #259, opened 2022-01-14
loschmidt.tilted_square_lattice gen+run

mpharrigan:
Following #236 , add scripts for generating a QuantumExecutableGroup and a script for running on a simulated EngineProcessor. 

This PR includes a very simple placeholder bash script for doing an "end-to-end" test that will be replaced with a more comprehensive solution when we have all the end-to-end parts committed. 

This PR does not include the data analysis steps, which will follow.

This requires https://github.com/quantumlib/Cirq/pull/4842

### DIFF
--- a/recirq/algorithmic_benchmark_library.py
+++ b/recirq/algorithmic_benchmark_library.py
@@ -1,7 +1,7 @@
 import dataclasses
 from dataclasses import dataclass
 from functools import lru_cache
-from typing import Type, Callable, List, Tuple, Any
+from typing import Type, Callable, List, Tuple, Any, Optional
 
 import pandas as pd
 
@@ -48,6 +48,7 @@ class AlgorithmicBenchmark:
     spec_class: Type['ExecutableSpec']
     executable_generator_func: Callable[[Any], 'QuantumExecutableGroup']
     configs: List['BenchmarkConfig']
+    description: Optional[str] = None
 
     def as_strings(self):
         """Get values of this class as strings suitable for printing."""
@@ -73,6 +74,7 @@ class BenchmarkConfig:
     full_name: str
     gen_script: str
     run_scripts: List[str]
+    description: Optional[str] = None
 
 
 BENCHMARKS = [
@@ -80,12 +82,28 @@ BENCHMARKS = [
         domain='recirq.otoc',
         name='loschmidt.tilted_square_lattice',
         executable_family='recirq.otoc.loschmidt.tilted_square_lattice',
+        description='\n'.join([
+            "An OTOC-style Loschmidt Echo on a tilted square lattice topology.",
+            "",
+            "This benchmark involves running a random unitary U forwards and backwards to",
+            "measure the fraction of times one ends up back in the starting state.",
+            "",
+            "Please browse the docstring for `TiltedSquareLatticeLoschmidtSpec` for details",
+            "on available parameters."
+        ]),
         spec_class=TiltedSquareLatticeLoschmidtSpec,
         executable_generator_func=get_all_tilted_square_lattice_executables,
         configs=[
             BenchmarkConfig(
                 short_name='small-v1',
                 full_name='loschmidt.tilted_square_lattice.small-v1',
+                description='\n'.join([
+                    "A 'small' configuration for quick verification of Loschmidt echos",
+                    "",
+                    "This configuration uses small grid topologies (making it suitable for",
+                    "running on simulators) and a small number of random instances making it",
+                    "suitable for getting a quick reading on processor performance in ~minutes."
+                ]),
                 gen_script='gen-small-v1.py',
                 run_scripts=['run-simulator.py']
             )
--- a/recirq/algorithmic_benchmark_library.py
+++ b/recirq/algorithmic_benchmark_library.py
@@ -1,7 +1,7 @@
 import dataclasses
 from dataclasses import dataclass
 from functools import lru_cache
-from typing import Type, Callable, List, Tuple, Any, Optional
+from typing import Type, Callable, List, Tuple, Any, Optional, Dict
 
 import pandas as pd
 
@@ -50,7 +50,7 @@ class AlgorithmicBenchmark:
     configs: List['BenchmarkConfig']
     description: Optional[str] = None
 
-    def as_strings(self):
+    def as_strings(self) -> Dict[str, str]:
         """Get values of this class as strings suitable for printing."""
         ret = {k: str(v) for k, v in dataclasses.asdict(self).items()}
         ret['spec_class'] = self.spec_class.__name__
--- a/recirq/algorithmic_benchmark_library.py
+++ b/recirq/algorithmic_benchmark_library.py
@@ -83,6 +83,12 @@ BENCHMARKS = [
         spec_class=TiltedSquareLatticeLoschmidtSpec,
         executable_generator_func=get_all_tilted_square_lattice_executables,
         configs=[
+            BenchmarkConfig(
+                short_name='small-v1',
+                full_name='loschmidt.tilted_square_lattice.small-v1',
+                gen_script='gen-small-v1.py',
+                run_scripts=['run-simulator.py']
+            )
         ]
     ),
 ]
--- a/recirq/cirqflow/run_utils.py
+++ b/recirq/cirqflow/run_utils.py
@@ -24,6 +24,9 @@ def get_unique_run_id(fmt: str = 'run-{i}', base_data_dir: str = '.') -> str:
         fmt: A format string containing {i} and optionally {date} to template trial run_ids.
         base_data_dir: The directory to check for the existence of files or directories.
     """
+    if '{i}' not in fmt:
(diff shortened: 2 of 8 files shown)


recirq/algorithmic_benchmark_library.py
```diff
@@ -83,6 +83,12 @@ class BenchmarkConfig:
         spec_class=TiltedSquareLatticeLoschmidtSpec,
         executable_generator_func=get_all_tilted_square_lattice_executables,
         configs=[
+            BenchmarkConfig(
+                short_name='small-v1',
```
dstrain115, line 87: Is there a good place to put a synopsis about what each benchmark is/does?  Either as a line comment or as a structured field?  Once we have a bunch of these, it's going to be more difficult to keep track of what each of them does.

recirq/cirqflow/run_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        fmt: A format string containing {i} and optionally {date} to template trial run_ids.
+        base_data_dir: The directory to check for the existence of files or directories.
+    """
+    if '{i}' not in fmt:
+        raise ValueError("run_id format string must contain an '{i}' placeholder.")
+
+    i = 1
+    while True:
+        run_id = fmt.format(i=i, date=datetime.date.today().isoformat())
+        if not os.path.exists(f'{base_data_dir}/{run_id}'):
```
dstrain115, line 33: Optional:  Not sure it matters much, but do you think it would be faster to list the directory and just check the results for each run rather than make O(n) separate os calls?  (Not sure how many runs we expect to do, but maybe could be a lot of them?)

recirq/otoc/loschmidt/tilted_square_lattice/end-to-end-test.sh
```diff
@@ -0,0 +1,3 @@
+python gen-small-v1.py
```
dstrain115, line 1: Worth a comment to explain this file?  Also, does this file need license info?

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
```
dstrain115, line 1: Add license info.

recirq/otoc/loschmidt/tilted_square_lattice/run-simulator.py
```diff
@@ -0,0 +1,24 @@
+import cirq
```
dstrain115, line 1: Add license info.

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
```diff
@@ -0,0 +1,22 @@
...
+
+import cirq
+from recirq.otoc.loschmidt.tilted_square_lattice import \
+    get_all_tilted_square_lattice_executables
+
+FN = 'loschmidt.tilted_square_lattice.small-v1.json.gz'
+
+
+def main():
```
dstrain115, line 10: Should add module level docstring to explain what this is and what the valid args and/or example command are.

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+
+import cirq
+from recirq.otoc.loschmidt.tilted_square_lattice import \
+    get_all_tilted_square_lattice_executables
+
+FN = 'loschmidt.tilted_square_lattice.small-v1.json.gz'
```
dstrain115, line 7: What is this?  Avoid abbreviations, especially when ambiguous (filename vs function).
Suggest OUTPUT_FILENAME or similar?

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def main():
+    exes = get_all_tilted_square_lattice_executables(
+        min_side_length=2, max_side_length=3, side_length_step=1,
+        n_instances=3,
+        macrocycle_depths=np.arange(0, 4 + 1, 1))
+    print(len(exes), 'executables')
+
+    cirq.to_json_gzip(exes, FN)
```
dstrain115, line 17: Should the output filename be a flag?

recirq/otoc/loschmidt/tilted_square_lattice/run-simulator.py
```diff
@@ -0,0 +1,24 @@
...
+from cirq_google.workflow import QuantumRuntimeConfiguration, \
+    SimulatedProcessorWithLocalDeviceRecord, execute
+from recirq.cirqflow.run_utils import get_unique_run_id
+from recirq.otoc.loschmidt.tilted_square_lattice import TiltedSquareLatticeLoschmidtSpec
+
+assert TiltedSquareLatticeLoschmidtSpec, 'register deserializer'
+
+FN = 'loschmidt.tilted_square_lattice.small-v1.json.gz'
```
dstrain115, line 9: Same comments about FN and docstrings in this file as well.

recirq/algorithmic_benchmark_library.py
mpharrigan: good idea, added

recirq/cirqflow/run_utils.py
mpharrigan: Yes, this is good to keep in mind. I'm imagining users will be coming up with meaningful run_id or run_id templates (particularly using the `{date}` option) so that we won't spend much time searching.

I'll punt on this since I think it would be straightforward to transparently update if and when it becomes a problem.

recirq/otoc/loschmidt/tilted_square_lattice/end-to-end-test.sh
mpharrigan: added.

Worth noting that I have much more robust plans for doing end-to-end (ie integration testing) once more of the benchmarks are in place; but I figured it wouldn't hurt to have this baby bash script for the time being.

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
mpharrigan: thanks

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
mpharrigan: changed.

also I think at some point this filename should be fetched maybe from the `algorithmic_benchmarking_library` "card catalog" module but I haven't completely thought through how that would work.

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
mpharrigan: added (using verbiage from the new description field in the algorithmic_benchmarking_library card catalog)

recirq/otoc/loschmidt/tilted_square_lattice/gen-small-v1.py
mpharrigan: The current design is to have a 1:1 correspondence between these generation scripts and the serialized `QuantumExecutableGroup`. I consider this file to be a configuration file like a `yaml` or `textproto` file. I could just use a yaml file or something; but I find often times the schema of configuration files gets extensions bolted on to it to handle new ways of configuring things so here I'm just using a real scripting language from the start.

Remember: we need to carefully control the benchmarks we run for accurate comparison over time. If you want to run an ad-hoc benchmark with your own parameter choices, you should likely create a new script to do what you want.

recirq/otoc/loschmidt/tilted_square_lattice/run-simulator.py
mpharrigan: thanks

recirq/otoc/loschmidt/tilted_square_lattice/run-simulator.py
mpharrigan: done

mpharrigan:
thanks for the review @dstrain115, PTAL

recirq/algorithmic_benchmark_library.py
```diff
@@ -48,6 +48,7 @@ class AlgorithmicBenchmark:
     spec_class: Type['ExecutableSpec']
     executable_generator_func: Callable[[Any], 'QuantumExecutableGroup']
     configs: List['BenchmarkConfig']
+    description: Optional[str] = None
 
     def as_strings(self):
```
dstrain115, line 53: Not related, but noticed it was missing return type.

dstrain115 (approved):

recirq/algorithmic_benchmark_library.py
mpharrigan: done

PR #260, opened 2022-01-29
Split up QuantumBoard.do_move()

losos0:
Split up a very long function to make development easier.

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -728,26 +728,21 @@ class CirqBoard:
         return result
 
     def do_move(self, m: move.Move) -> int:
-        """Will call do_move_internal to perform a move on the quantum board,
-        resetting the quantum circuit if the board is in a fully classical position.
-        """
-        initial_circuit = self.circuit.copy()
-        move_succeeded = self.do_move_internal(m)
-        if self.reset_starting_states and initial_circuit != self.circuit:
-            self.reset_if_classical()
-        return move_succeeded
-
-    def do_move_internal(self, m: move.Move) -> int:
         """Performs a move on the quantum board.
         Based on the type and variant of the move requested,
         this function augments the circuit, classical registers,
-        and post-selection criteria to perform the board.
+        and post-selection criteria to perform the move. The
+        circuit may be cleared if reset_if_classical is enabled and running the
+        circuit (which occurs only with that option enabled) shows that
+        the board is in a classical state.
+
         Returns:  The measurement that was performed, or 1 if
             no measurement was required.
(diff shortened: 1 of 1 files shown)


recirq/quantum_chess/quantum_board.py
```diff
@@ -728,31 +728,35 @@ def post_select_on(
...
-        if not m.move_type:
-            raise ValueError("No Move defined")
-        if m.move_type == enums.MoveType.NULL_TYPE:
-            raise ValueError("Move has null type")
-        if m.move_type == enums.MoveType.UNSPECIFIED_STANDARD:
-            raise ValueError("Move type is unspecified")
-
-        # Reset accumulations here because function has conditional return branches
+        move_func = {
+            enums.MoveType.JUMP: self._do_move_jump,
```
dstrain115, line 743: It would be nice if we could make this a constant somehow.  It seems like this is basically a static dictionary if not for the "self.".

recirq/quantum_chess/quantum_board.py
```diff
@@ -775,624 +779,599 @@ def do_move_internal(self, m: move.Move) -> int:
...
-            else:
-                raise ValueError(f"Invalid en passant target {m.target}")
-            epqubit = bit_to_qubit(epbit)
+        measurement = move_func[m.move_type](m, sbit, tbit, squbit, tqubit)
+        assert measurement in (0, 1)
+        if self.reset_starting_states and initial_circuit != self.circuit:
+            self.reset_if_classical()
+        return measurement
+
+    def _do_move_pawn_ep(self, m, sbit, tbit, squbit, tqubit):
```
dstrain115, line 788: We should add type info to this function and others.  Docstrings would be nice too.

recirq/quantum_chess/quantum_board.py
losos0: Done.

recirq/quantum_chess/quantum_board.py
losos0: Do you really think it would help? For me the key to understanding these functions is to read the place in do_move where the function table is used; having type and description boilerplate on each one wouldn't really help.

dstrain115: It would help in typechecking the contents of the function I suppose.  It's also possible to change the function table to be typed as a Dict[str,Callable[arg,types,here]] so that you could make sure that every function in the table has the same type.  

I think docstrings would be nice, especially for complicated functions so that various implementation details (like when  qubits are unhooked or whatever can be explained).  Also nice to have something to explain that "_ep" means en passant, for instance.

PR #265, opened 2022-02-06
Small quantum chess fixes

losos0:
- Use a smaller sample size in one place
- Make the code pass pytype checking

### DIFF
--- a/recirq/quantum_chess/controlled_iswap.py
+++ b/recirq/quantum_chess/controlled_iswap.py
@@ -14,6 +14,7 @@
 from typing import Sequence, Union, Optional
 
 import cirq
+from cirq import optimizers
 import cirq_google as cg
 import numpy as np
 from cirq.optimizers.two_qubit_to_fsim import (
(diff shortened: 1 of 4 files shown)


recirq/quantum_chess/controlled_iswap.py
```diff
@@ -116,8 +117,8 @@ def simplify_op(op: cirq.Operation) -> cirq.Operation:
     )
 
     converted_ops = cirq.Circuit(convert_op(op) for op in circuit.all_operations())
-    cirq.optimizers.MergeSingleQubitGates().optimize_circuit(converted_ops)
-    cirq.optimizers.DropNegligible().optimize_circuit(converted_ops)
+    optimizers.MergeSingleQubitGates().optimize_circuit(converted_ops)
```
dstrain115, line 120: Actually, can you remove the import above and change this to:

`cirq.MergeSingleQubitGates().optimize_circuit(converted_ops)`

This is safer, since we (cirq-maintainers) don't guarantee that package names will never change.  If we move this class around, the top-level import will still work.

dstrain115 (approved):

PR #266, opened 2022-02-16
Pytest skip breakage

mpharrigan:
Somehow this changed? Seeing errors in #259


verult (approved):
Not sure my approval is sufficient for merge but let's see!

PR #269, opened 2022-02-25
Fix f-strings in toric code

dstrain115:
- f'{blah=}' is not supported by the python runtime in
colab hosted runtimes.


mpharrigan:
Yes, we only run CI against python 3.9

PR #271, opened 2022-03-07
Fix import of now-private functions in QAOA

dstrain115:
- swap_rzz and rzz are now private functions.
- Update import statements as a workaround until a more
permanent fix can be done.


mpharrigan (approved):

mpharrigan:
oh man I was wondering why this was failing exactly where you aimed to fix it

dstrain115:
hmmm, looks like there are still some errors even after fixing the imports.  Something got screwed up with the quantum_chess imports.

PR #272, opened 2022-03-09
[loschmidt] plots.ipynb

mpharrigan:
This PR introduces `plots.ipynb`, which is a no-nonsense way of plotting data collected as part of the `loschmidt.tilted_square_lattice` benchmark. It is a stripped-down version of `analysis-walkthrough.ipynb`.

 - Factor out the aggregation operations explained in `analysis-walkthrough.ipynb` into the `analysis` module.
 - As part of the above, give more meaningful names to the intermediate dataframes.
 - For consistency and ease of visualizing multiple runs, plot errorbars in the vs_q_area plot instead of scatter points
 - somewhat unrelated change I'm sneaking in: use `str(processor_record)` instead of `processor_id` since SimulatedLocalProcessor's have the same processor_id as their real-life counterparts.

@verult @dstrain115

### DIFF
--- a/recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
+++ b/recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
@@ -133,7 +133,7 @@ def loschmidt_results_to_dataframe(results: ExecutableGroupResult) -> pd.DataFra
     return _results_to_dataframe(results, _to_record)
 
 
-def agg_vs_macrocycle_depth(df: pd.DataFrame):
+def agg_vs_macrocycle_depth(df: pd.DataFrame) -> Tuple[pd.DataFrame, List[str]]:
     """Aggregate mean and stddev vs macrocycle depth.
 
     We use pandas group-by functionality to
@@ -207,7 +207,7 @@ def fit_vs_macrocycle_depth(df):
     return fitted_df, exp_ansatz_vs_macrocycle_depth
 
 
-def agg_vs_q_area(df: pd.DataFrame):
+def agg_vs_q_area(df: pd.DataFrame) -> Tuple[pd.DataFrame, List[str]]:
     """Aggregate mean and stddev vs quantum area.
 
     quantum area is the circuit number of qubits multiplied by its depth.
--- a/recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
+++ b/recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
@@ -127,10 +127,42 @@ def loschmidt_results_to_dataframe(results: ExecutableGroupResult) -> pd.DataFra
             'instance_i': spec.instance_i,
             'n_repetitions': spec.n_repetitions,
             'success_probability': success_prob,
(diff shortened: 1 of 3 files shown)


recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
```diff
@@ -127,12 +127,44 @@ def _to_record(result: ExecutableResult,
...
             'n_repetitions': spec.n_repetitions,
             'success_probability': success_prob,
-            'processor_id': rt_config.processor_record.processor_id,
+            'processor_str': str(rt_config.processor_record),
         }
 
     return _results_to_dataframe(results, _to_record)
 
 
+def agg_vs_macrocycle_depth(df: pd.DataFrame):
```
dstrain115, line 136: Is it possible to define the return type here?

recirq/otoc/loschmidt/tilted_square_lattice/analysis.py
```diff
@@ -175,6 +207,41 @@ def _fit(row):
     return fitted_df, exp_ansatz_vs_macrocycle_depth
 
 
+def agg_vs_q_area(df: pd.DataFrame):
```
dstrain115, line 210: Return type here too.

dstrain115 (commented):
If I understand this, it looks like this is essentially plotting the results of analysis_walkthrough.ipynb.  Is this correct?

If so, I wonder if it makes more sense to just add this on as a section at the end of the analysis notebook.  This would flow better and also get rid of some of the redundant code.

What do you think?

mpharrigan:
> If so, I wonder if it makes more sense to just add this on as a section at the end of the analysis notebook. This would flow better and also get rid of some of the redundant code.

While there is some redundancy, I see this notebook as solving a completely different use case than the other one. The walkthrough is designed to give the developer some insight into the at times complicated aggregation routines available in the analysis module. It's necessary reading if you want to contribute a new way of slicing and dicing the data.

This other notebook is the final step in routine execution of the algorithmic benchmark. I must admit I have a preference for small, focused notebooks over big ones; but here are some real reasons:

 - limited noise (all the text describing the routines, describing the intermediate dataframe structures) in the plots notebook quickly lets the user focus on the plots and the results of the benchmark. They don't have to scroll all the way to the bottom of the other one
 - all the benchmarks will have a `plots.ipynb` since it's part of the normal user flow. Documentation-style notebooks will be added as necessary alongside if there's something that needs to be explained
 - everything in a notebook is essentially a global variable. There's a greater chance for unintended variable re-use in larger notebooks
 - in the future, I can imagine other/additional ways of displaying the resultant data than running the `plots` notebook by hand. 1) it can be run headlessly and upload the images somewhere 2) it can be copied into some actual dashboarding framework -- idk, but you definitely just want the couple-of-cells version for this

mpharrigan:
added type annotations

dstrain115 (approved):
Approved, with the caveat that we should explain which notebook to use somewhere (maybe in the overview doc?).  This can be in a later PR.

PR #278, opened 2022-03-30
Speed up analysis of board samples

losos0:
Take advantage of DataFrame vectorized operations to improve the
performance of CirqBoard.sample_with_ancilla(). (Measurement results are
returned by Cirq in a DataFrame.)

### DIFF
--- a/recirq/quantum_chess/quantum_board.py
+++ b/recirq/quantum_chess/quantum_board.py
@@ -16,6 +16,7 @@ from collections import defaultdict
 from typing import Dict, List, Optional, Sequence, Tuple
 
 import cirq
+import pandas
 
 from recirq.quantum_chess.bit_utils import (
     bit_to_qubit,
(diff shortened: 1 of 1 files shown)


recirq/quantum_chess/quantum_board.py
```diff
@@ -286,34 +286,37 @@ def sample_with_ancilla(
...
+            classical_state = self.state
+            for qubit in qubits:
+                if "anc" not in qubit.name:
+                    bit = qubit_to_bit(qubit)
+                    bitboard_exprs[bit // 32] += f"{qubit.name} * {1 << bit}.0 +"
+                    classical_state &= ~(1 << bit)
+            bitboard_exprs[0] += str(classical_state & (2**32 - 1)) + ".0"
+            bitboard_exprs[1] += str(classical_state >> 32 << 32) + ".0"
+            # Store evaluated position in bb_low and bb_high columns
+            data.eval("bb_low = {}\nbb_high = {}".format(*bitboard_exprs), inplace=True)
```
dstrain115, line 315: I am having trouble understanding what is going on here with all the bitwise operations.  It might be worth explaining this a bit more in some line comments.  For instance, explaining what the expected format of bitboard_exprs is.  Another alternative would be to put this into a helper function, and then you could explain it more fully in a docstring.

dstrain115 (approved):
Looks cool.  Added a note to add a bit more explanation for the bitboard construction.  Otherwise, it looks good to me.

PR #281, opened 2022-04-07
Install recirq in ftbbl.ipynb

dstrain115:
- This notebook is failing because recirq is not installed.
- Install it as part of the first cell.


dstrain115:
Merging to see if this fixes the notebook issues (which are broken anyway).  Might as well commit it and see if this fixes it overnight.

PR #283, opened 2022-04-12
Bump Cirq to 0.14

### LINKED ISSUE #247: Testing against Cirq 0.12
Following #237, we test against the current, previous, and next version of Cirq.

I've noticed as well that Cirq 0.12 does **not** include `cirq.NamedTopology`. With #236 this is a case where keeping compatibility with the prior version of Cirq is **not** worth it, esp. since v0.13 has been out for a while. I will disabled the "pre" check with a note to this issue. We'll re-enable when Cirq 0.14 comes out.

mpharrigan:
taking over from #277 

closes #247 

This should be relatively simple thanks to testing against `next` and by today's release of a compatible `pytket-cirq`.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         cirq-version:
           - 'current'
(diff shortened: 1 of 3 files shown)


dstrain115 (approved):
Nice, thanks for doing it correctly for me!

PR #284, opened 2022-04-19
Remove source viewing for ftbbl.ipynb

dstrain115:
- Try to remove source viewing to make ftbbl.ipynb
work with docs infrastructure.


dstrain115:
Merging since this notebook is broken anyway and I want to test it in the overnight build.

PR #285, opened 2022-04-28
Move beyond classical random circuits into ReCirq

dstrain115:
- These experiments probably belong in ReCirq and not in
      cirq.experiments.

See [cirq#3327](https://github.com/quantumlib/Cirq/issues/3327)

### DIFF
--- a/recirq/beyond_classical/__init__.py
+++ b/recirq/beyond_classical/__init__.py
@@ -1,0 +1,20 @@
+# Copyright 2022 The Cirq Developers
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


mpharrigan:
can we take this opportunity to change the module and function names to avoid "supremacy" and use "beyond classical"?

dstrain115:
> can we take this opportunity to change the module and function names to avoid "supremacy" and use "beyond classical"?

Done!  PTAL!

mpharrigan (approved):

mpharrigan:
Looks fine. I always find these circuits a little confusing because they correspond to the 2018 theory work https://www.nature.com/articles/s41567-018-0124-x and are subtly different from the circuits used in the experimental demonstration https://www.nature.com/articles/s41586-019-1666-5

dstrain115:
Thanks for the review.  We can revisit them and maybe add the newer circuit, but these are the ones from cirq.
This unblocks removing them from cirq, and we can always add more stuff (or extra explanation) to this directory now.

PR #286, opened 2022-05-02
Add extra imports to ftbbl.ipynb

dstrain115:
Add an import and install for cirq-google.

I am not totally sure why this fails to load routines from cirq_google, but I think it is cloning recirq, so it does not actually pip install during docs install.  Try adding explicit pip installs for cirq-google to fix broken imports on dev site.


mpharrigan:
Have you managed to test this from colab? Last I tried it was failing to install recirq at all -- seemingly because it couldn't find some packages on PyPI

dstrain115:
Closing this in favor of Matt Harrigan's (better) fix to just save outputs.

PR #287, opened 2022-05-02
Ftb/Bl walkthrough notebook -- save outputs

mpharrigan:
This reverts #284 and simply saves the output in this notebook.

As far as I can tell, pytket[-cirq] stipulates that it requires python 3.8+ whereas the colab runtime is 3.7.


dstrain115 (approved):
SGTM, let's do it.

PR #295, opened 2022-05-31
Move grid_parallel_two_qubit_xeb to ReCirq

dstrain115:
Move cirq.experiments.grid_parallel_two_qubit_xeb into
recirq.benchmarks.xeb

Part of cirq issue 3775.  This will move this out of cirq so that
we can deprecate this and cirq.experiments.cross_entropy_benchmarking.


mpharrigan (approved):
ok

we have a _lot_ of implementations of XEB, but if this is just a move then it's at least not increasing the number

PR #298, opened 2022-06-09
Fix broken import (cirq_google); redirect hyperlinks to unitary.

cognigami:

PR #300, opened 2022-06-16
[chess] So long, and thanks for all the fish!

cognigami:
Quantum Chess' move to //unitary is complete.


cognigami:
By way of heads-up: I didn't have rights to remove the two references to quantum chess in .github/workflows:
 - black.yml
 -  pythonpackage.yml

dstrain115:
Those two files are in the repository, so you should be able to edit them and submit them as part of the PR.
https://github.com/quantumlib/ReCirq/tree/master/.github/workflows

What permission errors are you seeing?

mpharrigan:
Woot! Maybe we should leave the `recirq/quantum_chess` directory with a README pointing to the new location. We can also have an `__init__.py` that raises an `ImportError` telling you to use the new package

cognigami:
Thanks @dstrain115, my token didn't have workflow permissions. I've pushed the commit now. Just waiting on advice on merging then patching PR #298 in to this PR.

.github/workflows/pythonpackage.yml
```diff
@@ -50,12 +50,6 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
-      - name: Test with pytest
-        run: |
-          pip install pytest
-          # RECIRQ_CHESS_TEST_SEED: make quantum_chess tests determnistic
-          # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations
-          RECIRQ_CHESS_TEST_SEED=789 RECIRQ_IMPORT_FAILSAFE=y pytest -v
```
mpharrigan, line 58: this disables pytest for the whole package. remove the chess_test_seed but we still need the call to `pytest`

mpharrigan (changes requested):

mpharrigan:
@cognigami what do you think about my suggestion:

> Maybe we should leave the recirq/quantum_chess directory with a README pointing to the new location. We can also have an __init__.py that raises an ImportError telling you to use the new package

quantum_chess/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Quantum Chess no longer exists in this repo. See README.md for mre info.
+
+raise ImportError
```
mpharrigan, line 3: you need to `raise ImportError()` so it actually constructs the exception object.

You can also add a helpful message: `raise ImportError("quantum_chess moved to github.com/quantumlib/unitary/")`

.github/workflows/pythonpackage.yml
mpharrigan: @cognigami this is an important issue to fix before merge ^

mpharrigan: still missing call to `pytest`

quantum_chess/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Quantum Chess no longer exists in this repo. See README.md for mre info.
+
+raise ImportError("quantum_chess moved to https://github.com/quantumlib/unitary/tree/main/unitary/quantum_chess")
```
mpharrigan, line 3: this should be under `recirq/quantum_chess` where the previous `__init__.py` file was

quantum_chess/README.md
```diff
@@ -0,0 +1,6 @@
+Quantum Chess has graduated to its own repository. You can follow the work on
```
mpharrigan, line 1: this should move to `recirq/quantum_chess` where the code used to be

.github/workflows/pythonpackage.yml
```diff
@@ -53,9 +53,7 @@ jobs:
       - name: Test with pytest
         run: |
           pip install pytest
-          # RECIRQ_CHESS_TEST_SEED: make quantum_chess tests determnistic
-          # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations
-          RECIRQ_CHESS_TEST_SEED=789 RECIRQ_IMPORT_FAILSAFE=y pytest -v
```
mpharrigan, line 58: this is a call to `pytest` (don't be confused by the preceding environment variable declarations). We cannot have this removed. It tests the whole repository.

.github/workflows/pythonpackage.yml
```diff
@@ -53,9 +53,7 @@ jobs:
...
+
```
mpharrigan, line 56:
```suggestion
          # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations
          RECIRQ_IMPORT_FAILSAFE=y pytest -v
```

mpharrigan (changes requested):
we need to still have the tests enabled for the rest of the repository

docs/_toc.yaml
```diff
@@ -68,4 +68,4 @@ toc:
 
 - heading: "Rabi oscillations"
 - title: "Rabi Oscillation Experiment"
-  path: /cirq/experiments/benchmarks/rabi_oscillations
```
mpharrigan, line 71: bad merge?

mpharrigan:
This seems to revert the changes from #305 ?

cognigami:
I think I made a bad merge. Let's see if I can say the right prayers to the git gods to undo this.

mpharrigan (approved):

PR #301, opened 2022-06-16
Use explicit precision for testing unitaries

mpharrigan:
Following https://github.com/quantumlib/Cirq/pull/5426/

This skips changes to quantum_chess directory in light of #300


dabacon (approved):
LGTM

mpharrigan:
remaining failures are in quantum chess. per above

PR #303, opened 2022-06-21
Don't need to pin h5py any more

mpharrigan:
https://github.com/quantumlib/OpenFermion-PySCF/issues/60


dstrain115 (approved):
Fine with me assuming that the test failures are confirmed to be unrelated.

mpharrigan:
Test failures unrelated. This error would show up in one of the hfvqe notebooks, which I tested offline

PR #306, opened 2022-06-27
Changes for Cirq 0.15

mpharrigan:
Fixes #304 

Fixes other things as well

PR #308, opened 2022-07-01
Change Result to ResultDict

dstrain115:
- Result instantiation is deprecated.  It is now an interface.
- Change references to ResultDict instead.

### DIFF
--- a/recirq/engine_utils.py
+++ b/recirq/engine_utils.py
@@ -174,7 +174,7 @@ class ZerosSampler(work.Sampler):
             for _, m, _ in meas:
                 assert len(m.qubits) == 1
             results = [
-                study.Result(
+                study.ResultDict(
                     params=p,
                     measurements={gate.key: np.zeros(
                         (repetitions, 1), dtype=int)
@@ -187,7 +187,7 @@ class ZerosSampler(work.Sampler):
             n_qubits = len(op.qubits)
             k = gate.key
             results = [
-                study.Result(
+                study.ResultDict(
                     params=p,
                     measurements={k: np.zeros(
                         (repetitions, n_qubits), dtype=int)})


dstrain115:
Ok, I think I got both pytest next and previous working now!

mpharrigan:
nice. probably some overlap with #306 fyi :)

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice.py
```diff
@@ -108,7 +108,10 @@ def _json_namespace_(cls):
         return 'recirq.otoc'
 
     def _json_dict_(self):
-        return dataclass_json_dict(self, namespace=self._json_namespace_())
+        if cirq.__version__ < '0.15':
+            return dataclass_json_dict(self, namespace=self._json_namespace_())
```
mpharrigan, line 112: is this necessary? I think the `_json_namespace_` was already present in 0.14 and the shim was for older compatibility

recirq/qaoa/placement.py
```diff
@@ -49,14 +49,14 @@ def calibration_data_to_graph(calib_dict: cg.Calibration) -> nx.Graph:
     return err_graph
 
 
-def _qubit_index_edges(device):
+def _qubit_index_edges(device: cirq.Device):
     """Helper function in `_device_to_tket_device`"""
-    dev_graph = ccr.gridqubits_to_graph_device(device.qubit_set())
+    dev_graph = ccr.gridqubits_to_graph_device(device.metadata.qubit_set())
```
mpharrigan, line 54: doesn't this give typing warnings about `metadata` might be `None`?

mpharrigan (approved):

mpharrigan:
We should likely merge this in favor of #306 because it is more comprehensive. It seems you have kept compatibility with Cirq 0.13 which is admirable but unnecessary. 

We need to bump the version spec for testing in `dev_tools/write-ci-requirements.py` so that current->0.15 and previous->0.14. That can be done in this PR if you want to relax some of the 0.13 shims or I can do it in a follow-on if you want to merge as-is.

recirq/otoc/loschmidt/tilted_square_lattice/tilted_square_lattice.py
dstrain115: I'll keep it for now to support 0.13 compatibility.  Feel free to remove it after you upgrade the requirements.

recirq/qaoa/placement.py
dstrain115: Done.

PR #309, opened 2022-07-10
GridDevice migration

verult:
* Migrated `device.qubits` callsites
  * A None check is added if `device` is a `cirq.Device`.
* Migrated 
* Removed gateset parameters

The first commit conflicts with https://github.com/quantumlib/ReCirq/pull/306 in that the solution is to check for `device.metadata is None` rather than requiring `GridDevice`. Either approach SGTM.

Couldn't test locally because `pytket.routing` is missing. @mpharrigan does https://github.com/quantumlib/ReCirq/pull/306 fix this?

### DIFF
--- a/recirq/benchmarks/rep_rate/rep_rate_calculator.py
+++ b/recirq/benchmarks/rep_rate/rep_rate_calculator.py
@@ -131,10 +131,11 @@ class RepRateCalculator:
         self.print_log = ''
 
     @classmethod
-    def from_engine(cls,
-                    engine: cg.Engine,
-                    processor_id: str,
-                    gate_set: Optional[cg.SerializableGateSet] = None):
+    def from_engine(
+        cls,
+        engine: cg.Engine,
+        processor_id: str,
+    ):
         """
         Constructs a RepRateTester using a cirq_google.Engine object.
         Uses the device from the device specification from the API
@@ -145,14 +146,9 @@ class RepRateCalculator:
             gate_set: gate set to use.  Defaults to square root of iswap
             processor_id: Processor to test on.
         """
-        gate_set = gate_set or cg.SQRT_ISWAP_GATESET
-        device = engine.get_processor(processor_id).get_device(
-            gate_sets=[gate_set])
-        sampler = engine.sampler(processor_id=processor_id, gate_set=gate_set)
-        gate = cirq.ISWAP**0.5
-        if gate_set == cg.SYC_GATESET:
-            gate = cg.SYC
-        return cls(device, sampler, gate)
+        device = engine.get_processor(processor_id).get_device()
+        sampler = engine.sampler(processor_id=processor_id)
+        return cls(device, sampler, cg.SYC)
 
     def _flush_print_log(self) -> None:
         """Remove print log used for debugging."""
--- a/recirq/benchmarks/rep_rate/rep_rate_calculator.py
+++ b/recirq/benchmarks/rep_rate/rep_rate_calculator.py
@@ -119,10 +119,13 @@ class RepRateCalculator:
             sampler: Sampler to test for rep rate.
             gate: two-qubit gate to use for sample circuits.
         """
+        if device.metadata is None:
+            raise ValueError("Invalid device: device must contain metadata.")
+
         self.device = device
         self.sampler = sampler
         self.gate = gate
-        self.qubits = list(self.device.qubits)
+        self.qubits = sorted(self.device.metadata.qubit_set) if self.device.metadata is not None else []
 
         # log of print statements for testing
         self.print_log = ''
--- a/recirq/benchmarks/rep_rate/rep_rate_calculator.py
+++ b/recirq/benchmarks/rep_rate/rep_rate_calculator.py
@@ -14,6 +14,13 @@ import cirq_google as cg
 import numpy as np
 
 
+_VALID_ENTANGLING_GATES = [
+    cirq.CZ,
+    cirq.SQRT_ISWAP,
+    cg.SYC,
+]
+
+
 def _latency_circuit(qubits: List[cirq.Qid]) -> cirq.Circuit:
     """Circuit for measuring round-trip latency.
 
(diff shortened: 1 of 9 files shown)


verult:
`QuantumProcessor` can no longer be immutable now that there's a single sampler. Possible solutions:
1. Make the dataclass mutable
2. Remove the cache

I went with the less disruptive approach but either way SGTM

verult:
@mpharrigan how do you suggest we handle the `previous` presubmit failure that SerializableDevice.metadata doesn't exist yet?

mpharrigan:
The previous / current / next designations haven't been updated to reflect the cirq 0.15 release, so hang tight

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -119,19 +119,23 @@ def __init__(self, device: cirq.Device, sampler: cirq.Sampler,
...
             gate: two-qubit gate to use for sample circuits.
         """
+        if device.metadata is None:
+            raise ValueError("Invalid device: device must contain metadata.")
+
         self.device = device
         self.sampler = sampler
         self.gate = gate
-        self.qubits = list(self.device.qubits)
+        self.qubits = sorted(self.device.metadata.qubit_set) if self.device.metadata is not None else []
```
mpharrigan, line 128: isn't the `if` redundant here because of the valueerror?

recirq/benchmarks/rep_rate/rep_rate_calculator.py
```diff
@@ -142,14 +146,9 @@ def from_engine(cls,
...
-        device = engine.get_processor(processor_id).get_device(
-            gate_sets=[gate_set])
-        sampler = engine.sampler(processor_id=processor_id, gate_set=gate_set)
-        gate = cirq.ISWAP**0.5
-        if gate_set == cg.SYC_GATESET:
-            gate = cg.SYC
-        return cls(device, sampler, gate)
+        device = engine.get_processor(processor_id).get_device()
+        sampler = engine.sampler(processor_id=processor_id)
+        return cls(device, sampler, cg.SYC)
```
mpharrigan, line 151: I think we still need to allow the user to specify `gate` since this changes the experiment. It's not just an argument to get an enginesampler

recirq/engine_utils.py
```diff
@@ -236,19 +227,13 @@ class QuantumProcessor:
...
-    _cached_samplers: Dict[Union[None, str], cirq.Sampler] \
-        = field(default_factory=dict)
-    _get_sampler_func: Callable[['QuantumProcessor', str], cirq.Sampler] = None
+    _cached_sampler: Optional[cirq.Sampler] = None
+    _get_sampler_func: Callable[['QuantumProcessor'], cirq.Sampler] = None
 
-    def get_sampler(self, gateset: str = None):
-        """Why must gateset be supplied?
-
-        https://github.com/quantumlib/Cirq/issues/2819
```
mpharrigan, line 246: heyo!

recirq/engine_utils.py
```diff
@@ -226,7 +217,7 @@ async def run_async(self, program: 'cirq.Circuit',
         return results[0]
 
 
-@dataclass(frozen=True)
+@dataclass
```
mpharrigan, line 220: Ah I see. We were cheating before because dataclass(frozen=True) can't enforce immutability if you sneak a dictionary into it. If we cared about this class, I'd probably migrate it differently; but new code should use `cg.ProcessorRecord` subclasses if you need a frozen dataclass serializable record of what processor you're using

recirq/benchmarks/rep_rate/rep_rate_calculator.py
mpharrigan: @dstrain115 what do you think

mpharrigan (commented):
lgtm except for the `gate` thing. @dstrain115 to opine

verult:
@mpharrigan just a heads up deprecation removals are under way for the device refactor, and SerializableDevice and SerializableGateSet are next to be removed. I'll update this PR soon to keep CI in a sane state :)

recirq/benchmarks/rep_rate/rep_rate_calculator.py
verult: I thought that as well, but mypy still complained...

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: Yeah, I think we want to specify the two-qubit gate here.
For instance, syc and sqrt-iswap have different compilation paths into waveforms and have different durations, so they (theoretically) could have different rep rates.  Also, SYC may not be supported on all processors.

If anything, I think we should default to CZ going forward, not SYC.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
dstrain115: I think it is getting confused between self.device and device.  I bet if you reorder things a bit, mypy won't complain.

dstrain115 (commented):
This seems fine to me once the tests pass and the gate parameter is added.

recirq/benchmarks/rep_rate/rep_rate_calculator.py
verult: aha that was it, thanks for the catch

recirq/benchmarks/rep_rate/rep_rate_calculator.py
verult: Updated

mpharrigan (approved):

mpharrigan:
If you merge in master, the CI will run against 0.15 as "current" and 1.0 as "next"

mpharrigan:
I merged master for you and fixed the conflicts

verult:
The failures in `next` seem to be due to API removals in 1.0

mpharrigan (approved):

mpharrigan:
force-merging to get us closer to 1.0 land

PR #314, opened 2022-07-20
Fix QAOA Maxcut Notebook for 1.0

dstrain115:
- Notebook had working_device.qubits which is the old style.
- Update for Device API.


augustehirth (approved):
Good catch, ty.

mpharrigan:
potentially redundant with #309 ?

PR #316, opened 2022-08-08
Fix some notebooks in Recirq

dstrain115:
- Used pytket.routing which was removed in pytket 1.0
- Fix use of deprecated cirq.optimizers.


mpharrigan (approved):
Thank you!

PR #317, opened 2022-08-10
Update docs with more Cirq 1.0 fixes

dstrain115:
- Hartree-Fock assumes str(q) == '(a, b)' but it is now 'q(a, b)'
- Remove TwoQubitGate
- Add metadata.qubit_set for device lookups.
- Add json_namespace to a few places.

### DIFF
--- a/docs/qaoa/hardware_grid_circuits.ipynb
+++ b/docs/qaoa/hardware_grid_circuits.ipynb
@@ -577,7 +577,7 @@
    "outputs": [],
    "source": [
     "from cirq.contrib.routing import gridqubits_to_graph_device\n",
-    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').qubit_set())\n",
+    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').metadata..qubit_set())\n",
     "nx.draw_networkx(device_graph, pos={q: (q.row, q.col) for q in device_graph.nodes}, node_color=QRED)"
    ]
   },
@@ -590,7 +590,7 @@
    "outputs": [],
    "source": [
     "from matplotlib import pyplot as plt\n",
-    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').qubit_set())\n",
+    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').metadata.qubit_set())\n",
     "matcher = nx.algorithms.isomorphism.GraphMatcher(device_graph, problem.graph)\n",
     "\n",
     "# There's a \"rotational\" freedom which we remove here:\n",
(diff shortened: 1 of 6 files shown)


docs/qaoa/hardware_grid_circuits.ipynb
```diff
@@ -577,7 +577,7 @@
    "outputs": [],
    "source": [
     "from cirq.contrib.routing import gridqubits_to_graph_device\n",
-    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').qubit_set())\n",
+    "device_graph = gridqubits_to_graph_device(recirq.get_device_obj_by_name('Sycamore23').metadata..qubit_set())\n",
```
mpharrigan, line 580: `metadata..`?

recirq/fermi_hubbard/execution.py
```diff
@@ -127,6 +127,10 @@ def cirq_resolvers(cls) -> Dict[str, Optional[Type]]:
     def _json_dict_(self):
         return cirq.dataclass_json_dict(self)
 
+    @classmethod
+    def _json_namespace_(cls):
+        return 'recirq'
```
mpharrigan, line 132: isn't this a breaking change? if we have experimental data saved without the recirq namespace?

docs/qaoa/hardware_grid_circuits.ipynb
dstrain115: Fixed.

recirq/fermi_hubbard/execution.py
dstrain115: Hmmm... I am not sure.  Is there a way to resolve this somehow?  The json code requires a namespace since it is not a cirq class.  Would we have to put a custom resolver into cirq, or can we somehow put a custom resolver into reCirq?

Have we encountered this before?

mpharrigan: well we've never changed the requirements for 3rd party code before (here: the change is we need a namespace when that used to not be required)

there's already a custom resolver to resolve this class's name (without the namespace). You're right: we could fix the issue by resolving both the namespaced and un-namespaced version for backwards compatibility

dstrain115: "we could fix the issue by resolving both the namespaced and un-namespaced version for backwards compatibility".  Do you know how to do this?  Can you point me in the right direction?

We probably should get this in, since I think there is a new batch of doc errors that is popping up, sigh...

mpharrigan: looks like @mrwojtek has put a `cirq_resolvers` classmethod on each of these objects that includes itself. In addition to `cls.__name__: cls` you'd add `f'recirq.{cls.__name__}: cls`

dstrain115: Done.  I will commit this shortly so I can start to make progress on the doc failures in recirq.

PR #322, opened 2022-09-12
Use Cirq's routing procedure instead of tket in qaoa experiments

ammareltigani:
Change the routing process in `/qaoa/placement.py` to use Cirq's new circuit routing implementation instead of depending on TKET.


tanujkhattar (commented):
This looks good to me. The two failing tests are because the current and previous released versions of Cirq don't have the RouteCQC method on them. 

cc @mpharrigan Can you please take a look? Does this change needs to be blocked on releasing a Cirq version that has the new RouteCQC method?

dstrain115:
Full review coming in a bit, but I strongly suggest that we put this change behind a flag or condition it on imports, since we want to support the current version (1.0) and ideally the previous version (0.15).

Putting this behind a conditional (if [cirq_routing_functions_exist] use cirq else use tket) will allow users more flexibility and a smooth upgrade path.

recirq/qaoa/placement.py
```diff
@@ -57,36 +33,6 @@ def calibration_data_to_graph(calib_dict: cg.Calibration) -> nx.Graph:
     return err_graph
 
 
-def _qubit_index_edges(device: cirq.Device):
```
dstrain115, line 60: We probably want to keep this, but mark it deprecated once cirq releases catch up.

recirq/qaoa/placement.py
```diff
@@ -106,25 +52,9 @@ def place_on_device(circuit: cirq.Circuit,
         initial_map: Initial placement of qubits
         final_map: The final placement of qubits after action of the circuit
     """
-    tk_circuit = pytket.extensions.cirq.cirq_to_tk(circuit)
```
dstrain115, line 109: I suggest using a flag or two different functions here for tket and cirq, to support backwards compatibility.

recirq/qaoa/placement.py
ammareltigani: ok!

mpharrigan:
OOO for now; but a feature flag would be nice

recirq/qaoa/placement.py
```diff
@@ -10,28 +10,39 @@
...
 try:
-    # Set the 'RECIRQ_IMPORT_FAILSAFE' environment variable to treat PyTket as an optional
-    # dependency. We do this for CI testing against the next, pre-release Cirq version.
-    import pytket
-    import pytket.extensions.cirq
-    from pytket.circuit import Node, Qubit
-    from pytket.passes import SequencePass, RoutingPass, PlacementPass
-    from pytket.predicates import CompilationUnit, ConnectivityPredicate
+    from cirq import RouteCQC
+    pytket = NotImplemented
```
ammareltigani, line 19: It's a little messy right now. I am declaring `pytket = NotImplemented` whenever Cirq's routing utilities exist to simplify the imports in `placement_test.py`. To remove this line I'd have to add another try-except block on the test file and change the `pytest.pskipif` code. Any opinions on which is better?

ammareltigani: I think the second option is cleaner so I pushed a new commit for it.

dstrain115 (approved):

PR #337, opened 2024-01-26
Fix serialization for reCirq objects

dstrain115:
- Somewhere along the line, the json_serializable_dataclass stoped working.
- This is because the functions need to be classmethods or else TaskClass._json_namespace_() fails (requiring positional argument obj)


mpharrigan (approved):
thanks!

PR #339, opened 2024-02-06
Fix serialization support for dataclasses

pavoljuhas:
`TestClass._json_dict_()` needs to be an instance method.

Fixes serialization of ReadoutScanTask in data_collection.ipynb notebook.

### DIFF
--- a/recirq/serialization_utils.py
+++ b/recirq/serialization_utils.py
@@ -213,8 +213,9 @@ def json_serializable_dataclass(_cls: Optional[Type] = None,
 
         cls._json_namespace_ = classmethod(lambda obj: namespace)
 
-        cls._json_dict_ = classmethod(lambda obj: cirq.obj_to_dict_helper(
-            obj, [f.name for f in dataclasses.fields(cls)]))
+        cls._json_dict_ = lambda obj: cirq.obj_to_dict_helper(
+            obj, [f.name for f in dataclasses.fields(cls)]
+        )
 
         if registry is not None:
             if namespace is not None:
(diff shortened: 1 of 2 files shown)


recirq/serialization_utils.py
```diff
@@ -213,8 +213,9 @@ def wrap(cls):
 
         cls._json_namespace_ = classmethod(lambda obj: namespace)
 
-        cls._json_dict_ = classmethod(lambda obj: cirq.obj_to_dict_helper(
-            obj, [f.name for f in dataclasses.fields(cls)]))
+        cls._json_dict_ = lambda obj: cirq.obj_to_dict_helper(
```
dstrain115, line 216: Are you sure, this is what I changed here:
https://github.com/quantumlib/ReCirq/pull/337
and it failed with the message "requiring positional argument obj"

pavoljuhas: I am looking at https://quantumai.google/cirq/dev/serialization#adding_a_new_serializable_value and to my understanding `_json_dict_` needs to be an instance method to return the instance data.

I left the `_json_namespace_` a classmethod - as it was set in your PR.  I have also updated the `test_json_serializable_class` unit test which fails at master, but passes in this PR.  Finally, I did a quick test run of the failing notebooks data_collection.ipynb and data_analysis.ipynb which seem to work after this change.

dstrain115: Ah, yes, you are correct.  I got json_dict and json_namespace confused.

dstrain115 (approved):

PR #342, opened 2024-05-04
Add QCQMC code part 1

fdmalone:
Adds the top level modules from the qcqmc code repo:

1. data: a light module for creating the data/ folder structure which is assumed throughout the code.
2. utilities: various utility functions (I already know this is going to be rejected :))
3. afqmc_circuits. Cirq circuits implementing some circuit primitives for the AFQMC trial wavefunction.

Also adds the requirements.

Screenshot 2024-05-03 at 11 12 40 PM

PR #343, opened 2024-05-04
QCQMC Part 2

fdmalone:
Adds the quaff dependency. Quite a big commit.


fdmalone:
Depends on #342

fdmalone:
@wjhuggins, PTAL at the very minimal README / suggest something more to say. It's not really clear from a cursory scan of the code if there's a useful minimal example to show some useful features of the library.

fdmalone:
@mpharrigan PTAL

recirq/third_party/__init__.py
mpharrigan, line 1: I think `third_party/` is supposed to go at the repository root and not be a python module itself, i.e. no `__init__.py`. and then futz around with the requirements or PYTHONPATH to pick up the `quaff` package

third_party/quaff/README.md
```diff
@@ -0,0 +1,11 @@
+# QUAFF
+
+Quaff is a Python library for manipulating, optimizing, and compiling Clifford gates.
```
fdmalone, line 3: @wjhuggins PTAL at this and make sure it's accurate please

fdmalone:
@mpharrigan PTAL I migrated it to a top level third_party folder and added a note in the README about the separate installation.

fdmalone:
cc @dstrain115 we decided adding the dependency to third_party was probably the best solution

third_party/quaff/README.md
wjhuggins: LGTM

fdmalone:
I cannot merge this myself

.github/workflows/pythonpackage.yml
```diff
@@ -50,11 +51,17 @@ jobs:
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
+      - name: Test third_party 
+        run: |
+          pip install pytest
+          pytest third_party/quaff
```
mpharrigan, line 57: you can probably remove this in a following PR when we have tests that exercise the quaff code path. We don't usually run tests for external packages

mpharrigan (approved):

PR #344, opened 2024-05-29
Change link from arXiv to citation of published paper

eliottrosenberg:
(Edited) Add citation for final published version of the KPZ paper.


dstrain115 (approved):
Please provide a description when you submit the PR that includes that this was the KPZ paper.

PR #351, opened 2024-06-20
QCQMC 6: Add experimental BluePrint

fdmalone:
Need to move to attrs


recirq/qcqmc/blueprint.py
```diff
@@ -0,0 +1,318 @@
...
+    Args:
+        circuit: The circuit to optimize
+
+    Returns:
+        The gate optimized circuit.
+    """
+
+    circuit = cirq.expand_composite(circuit)
+    circuit = cirq.align_left(circuit)
+
```
dstrain115, line 39: nit: weird newline here.

recirq/qcqmc/blueprint.py
```diff
@@ -0,0 +1,318 @@
...
+    return circuit
+
+
+def _to_tuple_of_tuples(
+    x: Iterable[Iterable[cirq.Qid]],
+) -> Tuple[Tuple[cirq.Qid, ...], ...]:
+    return tuple(tuple(_) for _ in x)
+
+
+def _to_tuple(x: Iterable[cirq.Circuit]) -> Tuple[cirq.Circuit, ...]:
```
dstrain115, line 51: This seems extraneous.  Why not just use tuple()?

recirq/qcqmc/blueprint.py
```diff
@@ -0,0 +1,318 @@
...
+    @property
+    def qubits(self) -> Tuple[cirq.Qid, ...]:
+        """The cirq qubits."""
+        return tuple(itertools.chain(*self.qubit_partition))
+
+    def _json_dict_(self):
+        return attrs.asdict(self)
+
+
+def build_blueprint(
```
dstrain115, line 291: Should this be a classmethod on BlueprintData?  such as BlueprintData.build_from_dependencies(..)

dstrain115 (approved):

recirq/qcqmc/blueprint.py
fdmalone: Maybe, this pattern of the factory methods being standalone methods separate from the classes is used elsewhere in the code too but I'm not entirely sure why...

recirq/qcqmc/blueprint.py
fdmalone: I think mypy complains about tuple as the converter similar to https://github.com/python/mypy/issues/8389 (not that it's really a problem since we're not type checking but my editor yells at me.

mpharrigan: type annotations in a dataclass probably

dstrain115: Oh, okay, let's put a comment in that this is needed for type conversion in the dataclass then.

recirq/qcqmc/blueprint.py
fdmalone: Done.

recirq/qcqmc/blueprint.py
fdmalone: Done.

recirq/qcqmc/blueprint.py
fdmalone: Done.

PR #353, opened 2024-06-28
Add QCQMC unit tests and speed up actions

fdmalone:
Add some unit tests for new files from #347 and speed up CI by setting OMP_NUM_THREADS=1. From experience elsewhere I **think** pyscf (or openblas) is considerably slower when this is picked up from the environment default (which is use all threads). Brings the CI times down from 13m ~ 6m which is what it was roughly before #347

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -55,7 +55,7 @@ jobs:
         run: |
           pip install pytest
           # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations
-          # EXPORT_OMP_NUMTHREADS: pyscf has poor openmp performace which slows down qcqmc tests.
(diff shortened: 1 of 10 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -55,6 +55,8 @@ jobs:
         run: |
           pip install pytest
           # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations
+          # EXPORT_OMP_NUMTHREADS: pyscf has poor openmp performace which slows down qcqmc tests.
```
dstrain115, line 58: nit performance

dstrain115 (approved):
LGTM.  Yes, usually CI has poor performance with multi-threading/processing for some reason.

.github/workflows/pythonpackage.yml
fdmalone: Done.

PR #354, opened 2024-06-29
QCQMC Part 7: Add Experiment classes

fdmalone:
Depends upon #351 and #353: Adds the Experiment class which packages everything together to run a simulated experiment.


recirq/qcqmc/experiment.py
```diff
@@ -0,0 +1,187 @@
...
+import attrs
+import cirq
+import numpy as np
+import qsimcirq
+from pytz import timezone
+
+from recirq.qcqmc import blueprint, config, data, for_refactor
+
+
+def _to_tuple(x: Optional[Iterable[float]]) -> Optional[Sequence[float]]:
```
dstrain115, line 27: I would consolidate this with other tuple functions or remove it.  This is simple enough we could add it inline, 
such as `tuple(x) if x is None else x`

recirq/qcqmc/experiment.py
```diff
@@ -0,0 +1,187 @@
...
+    raw_samples: np.ndarray = attrs.field(converter=np.asarray)
+    metadata: Dict[str, Any] = attrs.field(factory=dict)
+
+    def _json_dict_(self):
+        simple_dict = attrs.asdict(self)
+        simple_dict["params"] = self.params
+        return simple_dict
+
+
+def build_experiment(
```
dstrain115, line 94: Same here.  Add as a classmethod on experiment data?

recirq/qcqmc/experiment_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+import cirq
+
+from recirq.qcqmc.blueprint import BlueprintData
+from recirq.qcqmc.experiment import SimulatedExperimentParams, build_experiment
+from recirq.qcqmc.hamiltonian import HamiltonianData
+from recirq.qcqmc.trial_wf import TrialWavefunctionData
+
+
+def test_small_experiment_raw_samples_shape(
```
dstrain115, line 25: This test file seems short.  Does it test everything in experiments?

recirq/qcqmc/experiment.py
fdmalone: Done.

recirq/qcqmc/experiment.py
fdmalone: Done.

recirq/qcqmc/experiment_test.py
fdmalone: It does at least call every function (after a small update) but the test coverage is not great. I think I'd @wjhuggins to help writing some sensible physical tests of the shadow tomography protocol but I think that can be a follow up (we discussed writing some better documentation of the shadow tomography part of the code which is probably of most interest to QIS people).

dstrain115: Ok, yeah, let's make it a follow up PR.

dstrain115 (approved):

PR #361, opened 2024-07-29
Move third_party to recirq/third_party

fdmalone:
Should fix pip installation within notebooks.


recirq/third_party/quaff/basis_change_test.py
```diff
@@ -1,7 +1,7 @@
 import cirq
 import numpy as np
 import pytest
-import quaff
+from recirq.third_party import quaff as quaff
```
pavoljuhas, line 4: nit - `as quaff` is redundant -
```suggestion
from recirq.third_party import quaff
```
also please fix the same import in comb_test.py, cz_layer_test.py, indexing_test.py, linalg_test.py, random_test.py

recirq/third_party/quaff/json_resolver.py
```diff
@@ -4,9 +4,9 @@
 
 
 def json_resolver(quaff_type: str) -> Optional[ObjectFactory]:
-    if quaff_type.startswith("quaff."):
+    if quaff_type.startswith("quaff"):
         return cast(
             ObjectFactory,
-            eval(f"__import__('quaff').{quaff_type.split('.', 1)[1]}", {}, {}),
+            eval(f"__import__('recirq').third_party.{quaff_type}", {}, {}),
```
pavoljuhas, line 10: Please rewrite without eval, for example, after adding module-level import `from recirq.third_party import quaff` -
```suggestion
            getattr(quaff, quaff_type[6:]
```

recirq/third_party/quaff/json_resolver.py
```diff
@@ -4,9 +4,9 @@
 
 
 def json_resolver(quaff_type: str) -> Optional[ObjectFactory]:
-    if quaff_type.startswith("quaff."):
+    if quaff_type.startswith("quaff"):
```
pavoljuhas, line 7: The previous prefix `quaff.` seems more robust.
Is there some edge case that does not work with `quaff.` and needs `quaff` instead?

pavoljuhas (changes requested):
Few small suggestions, but otherwise LGTM.

pavoljuhas:
One more comment - after merging with master the high_level.ipynb notebook fails when trying to read `.../site-packages/recirq/qcqmc/data/integrals/fh_sto3g/hamiltonian.chk`.

Perhaps that file needs to be added to package_data here?
https://github.com/quantumlib/ReCirq/blob/17a5fdef73e0e578eb8df33b2c9f8418630ed88c/setup.py#L56-L61

recirq/third_party/quaff/basis_change_test.py
fdmalone: Whoops, thanks.

recirq/third_party/quaff/json_resolver.py
fdmalone: Thanks, done.

recirq/third_party/quaff/json_resolver.py
fdmalone: No good reason, reverted.

fdmalone:
@pavoljuhas PTAL I think I resolved everything and verified the notebooks run in a clean environment.

pavoljuhas (approved):

PR #364, opened 2024-07-31
Update NumPy APIs that will be removed in numpy-2.0

pavoljuhas:
Executed  ruff check --select NPY201 --fix ./

Ref: https://numpy.org/doc/stable/numpy_2_0_migration_guide.html


pavoljuhas:
@fdmalone - PTAL

fdmalone (approved):

PR #374, opened 2025-02-27
Add common repository files for GitHub best practices

mhucka:
This adds a few files like a code of conduct and security policy.

It also updates and enhances the CONTRIBUTING.md guidelines.


CONTRIBUTING.md
```diff
@@ -1,29 +1,123 @@
...
+use the tools provided by GitHub for [pull
+requests](https://help.github.com/articles/about-pull-requests/) for this
+purpose. The preferred manner for submitting pull requests is to fork the ReCirq
+[repository](https://github.com/quantumlib/ReCirq), create a [git
+branch](https://git-scm.com/book/en/v2/Git-Branching-Branches-in-a-Nutshell) in
+this fork to do your work, and when ready, create a pull request from this
+branch to the main ReCirq repository. The subsections below describe the process
+in more detail.
+
+Pleae make sure to follow the [Google Style
```
pavoljuhas, line 46:
```suggestion
Please make sure to follow the [Google Style
```

CONTRIBUTING.md
```diff
@@ -1,29 +1,123 @@
...
+    corner of the [repository page](https://github.com/quantumlib/ReCirq)).
+    Forking creates a new GitHub repo at the location
+    `https://github.com/USERNAME/ReCirq`, where `USERNAME` is your GitHub user
+    name.
+
+1.  Clone (using `git clone`) or otherwise download your forked repository to
+    your local computer, so that you have a local copy where you can do your
+    development work using your preferred editor and development tools.
+
+1.  Check out the `main` branch and create a new git branch from `main`:
```
pavoljuhas, line 62: The mainline branch in this repository is `master`, please update here and below.

Alternatively, you can switch the default branch to the recommended `main`, but that should be done in a separate PR.

Cirq had this switched earlier in https://github.com/quantumlib/Cirq/pull/6363.

CONTRIBUTING.md
```diff
@@ -1,29 +1,123 @@
...
+Do your work and `git commit` your changes to your branch as needed.
+
+We use several tools to test code and perform other activities such as checking
+formatting against the style guidelines. You can run those tools locally during
+development. Wrapper scripts are located in the [`check/`](./check/)
+subdirectory to simplify running the tools.
+
+*   Run `check/pytest` to run the Pytest suite
+*   Run `check/mypy` to run the Mypy type checker
+*   Run `check/pylint` to run the Pylint code linter
```
pavoljuhas, line 81: These do not exist in ReCirq.  Please update to reflect the actual testing tools available, the jobs in [pythonpackage.yml](https://github.com/quantumlib/ReCirq/blob/a8a5791c17df4592dd63a7b50ea67a447e7e5c96/.github/workflows/pythonpackage.yml#L45-L59) could be of some guidance.  Note the CI jobs are in a bit wonky state, for example, the flake8 is run as a soft check and reports most errors as warnings.

pavoljuhas (changes requested):
Please see inline comments in CONTRIBUTING.md.  Otherwise LGTM.

CONTRIBUTING.md
mhucka: You're right. Looking at the situation now, the tools that are being used need to be installed, but they're not in the requirements file. This is going to need a little bit more to make it user-friendly, so for the time being, I'm going to remove the explanations of the tools and just leave the first sentence in the "Development and testing" section, and open an issue about improving the tool setup.

dstrain115 (approved):

CONTRIBUTING.md
pavoljuhas: The instructions still refer to the `main` branch which does not exist.

Can you please comment if you plan on renaming `master` to `main` in this repo soon?
If so the text can stay as is.  Otherwise we should use `master` now and replace it with `main` later after the rename.

pavoljuhas (approved):
LGTM, but please check the `main` vs `master` comment.

CONTRIBUTING.md
mhucka: Ack, foo. Switching between projects has fried my ability to remember which ones use which. (And yes, I think eventually we should rename master to main, but not this week, so I'll fix this up instead.)

pavoljuhas (approved):

PR #385, opened 2025-04-30
Remove first-interaction workflow

mhucka:
It has not proved popular and instead has proved problematic.


mhucka:
Duplicate of #382.

PR #387, opened 2025-04-30
Replace nbfmt with nbformat

mhucka:
Previously, `dev_tools/nbfmt` downloaded a copy of `nbformat` from the Cirq repo, and also downloaded a requirements file so that it could install the TensorFlow docs Python package. This was not good from a software engineering and maintenance perspective.

The changes in this PR do the following:

- Replace `nbfmt` with `nbformat` copied from Cirq
- Introduce a subdirectory, `dev_tools/requirements/`, that follows the same scheme as Cirq, so that people working with both systems will find it familiar
- Add a requirements file to `dev_tools/requirements` for the notebook testing environment
- Updates the CI workflow in `.github/workflows` to work with the changes above

At the moment, `dev_tools/requirements/` only contains files related to notebook testing. Future PRs may move some of the other requirements files into there, to bring it closer to how Cirq is organized.


pavoljuhas (changes requested):
The script needs only the `tensorflow-docs` package to run.
Can we limit the installed requirements to just that one [as in cirq](https://github.com/quantumlib/Cirq/blob/de37106efe5f6b36e53620ea7319f53897b61991/.github/workflows/ci.yml#L142-L145)?

pavoljuhas (changes requested):
Let us make this PR focused on one task - fixing check/nbformat - only and leave any updates of check_notebooks.py dependencies to another PR.  `check/nbformat` needs just the `tensorflow-docs` package to run, which is the only one installed in cirq CI - https://github.com/quantumlib/Cirq/blob/575552a58f8c8c2550af4ddee49734ae57f60f3a/.github/workflows/ci.yml#L142-L145.

I'd suggest to copy and use only the [tensorflow-docs.txt](https://github.com/quantumlib/Cirq/blob/main/dev_tools/requirements/deps/tensorflow-docs.txt) from cirq here and remove any other new requirements files.  They can be added later on if and when they would be necessary.

Otherwise LGTM.

dev_tools/nbformat
```diff
@@ -0,0 +1,69 @@
...
+
+# Get the working directory to the repo root.
+thisdir="$(dirname "${BASH_SOURCE[0]}")" || exit $?
+topdir="$(git -C "${thisdir}" rev-parse --show-toplevel)" || exit $?
+cd "${topdir}" || exit $?
+
+# check if the nbfmt tool in tensorflow-docs is installed and working
+if ! python3 -c "import tensorflow_docs.tools.nbfmt"; then
+    echo -e "\033[31mMissing installation of tensorflow-docs Python package.\033[0m"
+    echo -e "\033[31mPlease do 'pip install -r dev_tools/requirements/notebook.env.txt'.\033[0m"
```
pavoljuhas, line 45: The previous wording of the error message from https://github.com/quantumlib/Cirq/pull/7004 might be more informative.  The problem fixed there was that the tensorflow-docs package _was installed_, but failed to import because of incompatibility of protobuf versions.

mhucka: > The previous wording of the error message from [quantumlib/Cirq#7004](https://github.com/quantumlib/Cirq/pull/7004) might be more informative. The problem fixed there was that the tensorflow-docs package _was installed_, but failed to import because of incompatibility of protobuf versions.

I personally found the message _less_ informative, or else I wouldn't have taken time to try to improve it. 

But your point is taken: the `import` can fail for reasons other than the tensorflow-docs package not being installed, so the new message may not always be correct.

It's unfortunate that improving this would take too much time. I guess I'll revert the message.

mhucka:
> Let us make this PR focused on one task - fixing check/nbformat - only and leave any updates of check_notebooks.py dependencies to another PR. `check/nbformat` needs just the `tensorflow-docs` package to run, which is the only one installed in cirq CI - [quantumlib/Cirq@`575552a`/.github/workflows/ci.yml#L142-L145](https://github.com/quantumlib/Cirq/blob/575552a58f8c8c2550af4ddee49734ae57f60f3a/.github/workflows/ci.yml#L142-L145).
> 
> I'd suggest to copy and use only the [tensorflow-docs.txt](https://github.com/quantumlib/Cirq/blob/main/dev_tools/requirements/deps/tensorflow-docs.txt?rgh-link-date=2025-05-05T23%3A38%3A46Z) from cirq here and remove any other new requirements files. They can be added later on if and when they would be necessary.
> 
> Otherwise LGTM.

The latest commits do this (hopefully).

pavoljuhas (approved):

PR #383, opened 2025-04-22
Adding ReCirq for Lattice Gauge Theory paper accepted to Nature

snichet:
In this Recirq Tutorial, I have outlined cirq simulations of the experiments that I conducted as a hardware resident at Google quantum AI. These results are outlined in the arXiv posting https://arxiv.org/abs/2409.17142.

PR #388, opened 2025-05-11
New PR Adding ReCirq for Lattice Gauge Theory paper accepted to Nature

snichet:
Accidentally deleted the fork for previous PR (https://github.com/quantumlib/ReCirq/pull/383). Submitting new PR.


mhucka:
Just a quick note that we're aware CI checks are currently failing for all PRs. I'm working on fixing that.

snichet:
@dstrain115 I've recently seen this error in other contexts as well. `cirq.Simulator()` get in an unhappy state, complaining that the state vector is not normalized when computing expectation values. Was there a recently a PR that would have changed the floating point precision an some point?

snichet:
As a work-around, I just switched `cirq.Simulator()` to `qsimcirq.QSimSimulator()` and all tests are now passing.

recirq/lattice_gauge/lattice_gauge_experiment.py
```diff
@@ -0,0 +1,872 @@
...
+                    axis=1,
+                )
+                % 2
+            )
+
+    return plaquette_bitstrings
+
+def cnot_on_layer(
+    pairs_list: Sequence[tuple[cirq.GridQubit, cirq.GridQubit]],
+    depolarization_probability: float | dict | None = None,
```
dstrain115, line 588: It looks like the colab is complaining about this.  It looks like you were executing your colab on python 3.9 which is over 5 years old and doesn't support this.  Can you either clear the output of the cell or rerun in python 3.10+?

```
/Users/tylerac/Documents/VS_Code_Projects/ReCirq Stuff/.venv/lib/python3.9/site-packages/urllib3/__init__.py:35:
...

    586 def cnot_on_layer(
    587     pairs_list: Sequence[tuple[cirq.GridQubit, cirq.GridQubit]],
--> 588     depolarization_probability: float | dict | None = None,
    589 ) -> Sequence[cirq.Moment]:
    590     """Outputs a list of moments for CNOT between two lists, in terms of CZ gates.
    591 
    592     Args:
   (...)
    596             dictionary that maps specific depolarization probability to specific pairs of qubits.
    597     """
    598     if depolarization_probability is None:

TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

recirq/lattice_gauge/lattice_gauge_experiment.py
```diff
@@ -0,0 +1,872 @@
...
+from typing import Any
+
+import cirq
+from matplotlib import colormaps
+import matplotlib.patches as mpatches
+import matplotlib.pyplot as plt
+import numpy as np
+import sympy
+
+from .lattice_gauge_grid import QubitNeighbor, LGTGrid
```
dstrain115, line 25: Prefer absolute imports.

dstrain115 (approved):
Aside from the errors in the colab output, it looks good to me.
I would just re-run the colab output and then we're good to go.

recirq/lattice_gauge/lattice_gauge_experiment.py
snichet: Oh, thanks for pointing this out @dstrain115 ! I actually didn't know that the outputs on the notebooks were pushed along with the input code cells. I see that in the Toric Code reCirq at https://quantumai.google/cirq/experiments/toric_code/toric_code_ground_state has all the cells evaluated, but when opened as a Colab it does not have all cells evaluated (more like a tutorial). However for the kpz reCirq, https://quantumai.google/cirq/experiments/kpz/kpz, both the quantumai site and the colab have all cells evaluated. Is there a preference if I should execute all of the cells or remove them all?

dstrain115: If you leave all the cells blank, then documentation infrastructure will evaluate them.  This is generally better since it will function as an integration test (ie. if a cirq upgrade breaks it, we'll get a notification).

However, if there are any calls to quantum engine (or other external services that need authentication) or if some of the cells take a long time, then it is better to evaluate it before-hand.  For instance, I see that the tutorial goes up to 20 qubits.  I am not sure how long that takes to run, but if it is longer than ~1 minute or so, it might be better to save the output rather than risk sporadic timeouts.

eliottrosenberg:
@snichet @dstrain115 is this ready to merge?

recirq/lattice_gauge/lattice_gauge_experiment.py
snichet: @dstrain115 on my laptop, I think the slowest cell took 45 seconds to execute. Thus I added a few disclaimers. See below:

image

snichet: This cell does not have an output, though. But to generate the plot in the next cell requires these objects.

snichet:
> @snichet @dstrain115 is this ready to merge?

almost there. just need to decide if we will leave the cells evaluated or not. do you have an opinion?

dstrain115:
I don't have a strong opinion, but maybe just leave the cell outputs so that we know it is good data and for simplicity's sake?

tcochran-quera:
@dstrain115 @eliottrosenberg @mhucka

Let's send it! Thanks for everyone's help :)

eliottrosenberg:
Thanks, Tyler! @dstrain115 do you want to merge it? (I don't have the ability to do that.)

mhucka:
I can merge it too, but I think @dstrain115 should have the last word.

dstrain115:
Merged.  Thanks everyone!

eliottrosenberg:
@snichet @tcochran-quera I know I'm a little late in pointing this out, but you probably also want to update this file to add your work to the table of contents on the website: https://github.com/quantumlib/ReCirq/blob/master/docs/_index_included.yaml.

PR #392, opened 2025-05-30
Add Lattice Gauge notebook to the TOC

dstrain115:
- This adds the Lattice Gauge Theory notebook to the devsite table of contents.

### DIFF
--- a/docs/_index_included.yaml
+++ b/docs/_index_included.yaml
@@ -133,5 +133,3 @@ included:
       path: /cirq/experiments/lattice_gauge/lattice_gauge
     - heading: Code Layout
 
-
-docs/lattice_gauge/lattice_gauge.ipynb
--- a/docs/_index_included.yaml
+++ b/docs/_index_included.yaml
@@ -123,3 +123,15 @@ included:
     - heading: Experimental Wavefunctions
       description: Analyze published experimental wavefunctions.
       path: /cirq/experiments/qcqmc/experimental_wavefunctions
+
+- heading: Lattice Gauge Theory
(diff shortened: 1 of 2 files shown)


dstrain115:
@snichet @eliottrosenberg to review as well (they are not contributors to reCirq so cannot be properly assigned).

eliottrosenberg (approved):
Thanks, @dstrain115!

snichet:
Thanks much @dstrain115, @mhucka, and @eliottrosenberg ! I believe the article appears in print on June 12th, but online June 4th. So it would be good to have the link available on June 4th. For reference I included the url in the references of the paper as https://quantumai.google/cirq/experiments/lattice_gauge/lattice_gauge. I hope this is correct

docs/_index_included.yaml
```diff
@@ -123,3 +123,15 @@ included:
...
+  description: Simulations for Charges and Strings in (2+1)D Lattice Gauge Theories
+  options:
+    - cards
+  items:
+    - heading: Experiment Example
+      path: /cirq/experiments/lattice_gauge/lattice_gauge
+    - heading: Code Layout
+
+
+docs/lattice_gauge/lattice_gauge.ipynb
```
pavoljuhas, line 137: This looks like a leftover and also breaks yaml syntax. Please delete.

pavoljuhas (approved):
LGTM after suggested clean up.

eliottrosenberg:
@dstrain115 I think this is in your court.

docs/_index_included.yaml
dstrain115: Done.

PR #402, opened 2025-07-31
Fix errors occuring with docs

dstrain115:
- Looks like a few ipynb files are failing to generate docs due to numpy 2.0 issues.
- Changed np.alltrue to np.all (np.alltrue was removed in 2.0)
- Casted type from uint8 to int8 due to overflows of -1.


mhucka (approved):
LGTM

PR #415, opened 2025-08-16
Open tutorial file using normal mode

dstrain115:
- The option 'r|xz' opens an lzma compressed stream for reading.
- This causes a StreamError: "seeking backwards is not allowed".
- This changes it to 'r:xz' which opens it as a normal file so it is seekable.
- This fixes a problem with rendering the documentation for reCirq

### DIFF
--- a/recirq/documentation_utils.py
+++ b/recirq/documentation_utils.py
@@ -71,7 +71,7 @@ def fetch_guide_data_collection_data(base_dir=None):
     # Read into a BytesIO object to make it seekable
     stream_bytes = io.BytesIO(stream.read())
 
-    with tarfile.open(fileobj=stream_bytes, mode='r|xz') as tf:
+    with tarfile.open(fileobj=stream_bytes, mode='r:xz') as tf:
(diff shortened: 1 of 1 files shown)


mhucka:
This is peculiar. When I run `pytest recirq/documentation_utils_test.py` on MacOS under Python 3.11, it passes, yet it does contain a test that invokes `fetch_guide_data_collection_data`. The test is evidently insufficient. @dstrain115 How did you get the stream error? We should improve the test.

dstrain115:
> This is peculiar. When I run `pytest recirq/documentation_utils_test.py` on MacOS under Python 3.11, it passes, yet it does contain a test that invokes `fetch_guide_data_collection_data`. The test is evidently insufficient. @dstrain115 How did you get the stream error? We should improve the test.

The function `fetch_guide_data_collection_data` downloads data from the internet, so the test patches `@mock.patch('urllib.request.urlopen')` so somehow this error doesn't manifest then.

I ran it in a colab.

PR #423, opened 2025-08-19
Fix #189: handle use_physical_cz argument

### LINKED ISSUE #189: Unexpected keyword argument in `build_otoc_circuits`

mhucka:
The call to `cz_to_sqrt_iswap` passed an argument that it doesn't have. Based on the description of `use_physical_cz`, it seems that the intention was to call `cirq.CZ` when `use_physical_cz=True`.


pavoljuhas (approved):
Before merging, please delete the `cz_correction` argument of `build_otoc_circuits` as it is unused.

PR #425, opened 2025-09-01
Fix #406: add missing security scan workflow

### LINKED ISSUE #406: Configure missing security scans
We should add the security-scanning workflows that we use in other repos: the OSV scanner, Scorecard, and a more complete CodeQL scan configuration.

mhucka:
This adds a workflow to run the open-source vulnerabilities (OSV) scanner that we run in other Quantumlib repos. It also updates the Scorecard scanner workflow to run on PRs, not just weekly, and also print info about the results to the job summary page. Finally, this renames `ossf-scorecard.yaml` to `scorecard-scanner.yaml` in an effort to make the purpose a little more clear.

### DIFF
--- a/.github/workflows/scorecard-scanner.yaml
+++ b/.github/workflows/scorecard-scanner.yaml
@@ -102,7 +102,7 @@ jobs:
       - name: Write the Scorecard report page link to the workflow summary
         run: |
           repo="${{github.repository}}"
-          url="https://scorecard.dev/viewer/?uri=github.com%2F${repo//\//%2F}"
+          url="https://scorecard.dev/viewer/?uri=github.com/${repo}"
           {
           echo -n "The results are available on the OpenSSF Scorecard "
           echo "[report page for ${{github.repository}}]($url)."


.github/workflows/scorecard-scanner.yaml
```diff
@@ -0,0 +1,109 @@
...
+  write-summary:
+    name: Scorecard results
+    needs: run-scorecard
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - name: Write the Scorecard report page link to the workflow summary
+        run: |
+          repo="${{github.repository}}"
+          url="https://scorecard.dev/viewer/?uri=github.com%2F${repo//\//%2F}"
```
pavoljuhas, line 105: For chrome the link works without extra quoting, e.g., for cirq at
https://scorecard.dev/viewer/?uri=github.com/quantumlib/Cirq

```suggestion
          repo="${{github.repository}}"
          url="https://scorecard.dev/viewer/?uri=github.com/${repo}"
```

Or does this break in some other browser?

pavoljuhas (approved):
LGTM, but please see minor comment.

.github/workflows/scorecard-scanner.yaml
mhucka: Well, due to being unable to test some of the other browsers (particularly Microsoft's), I thought it'd be safer to quote it. However, think about it some more, this is not a case where we need to be ultra-portable, and since it works unquoted for Chrome & Safari (and they account for ~80% of all browser use), the simpler form is good enough, and is certainly more readable. So, thanks for the suggestion.

Issue #229, opened 2021-11-10
Rename master branch to main


mpharrigan:
I'd like to rename the `master` branch to `main`, see https://github.com/github/renaming

mhucka:
Resolved by PR #440.

PR #440, opened 2025-09-29
Update references to 'master' to be to 'main' instead

mhucka:
This is part of addressing #229. This changes references to branch `master` to refer to `main` instead.


pavoljuhas (approved):
It seems that the recirq/readout_scan/Readout-Analysis.ipynb symlink was accidentally replaced with regular file.  Should we put it back?

Otherwise LGTM.

PS: Please also submit follow-up PRs to Cirq and pyle which have URL links to the ReCirq master branch.

PR #465, opened 2026-03-15
Use `-n auto` for pytest by default

mhucka:
This adds `-n auto` to the top-level `pytest.ini` file and removes the  `-n` flag from invocations in the CI workflows. (The latter is because pytest will pick it up from the `pytest.ini` file now.)

Using parallelism speeds up pytest execution by a massive factor. I encountered no problems with it in testing, and the CI check workflows was already using it, so it looks safe to use as a default.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -93,7 +93,7 @@ jobs:
           # OMP_NUM_THREADS: PySCF's poor OpenMP performance slows down QCQMC tests.
           # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations.
           export OMP_NUM_THREADS=1
-          RECIRQ_IMPORT_FAILSAFE=y pytest -n logical -v --skipslow
+          RECIRQ_IMPORT_FAILSAFE=y pytest -v --skipslow
 
   pylint:
     name: Run Python linters
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,6 @@
 [pytest]
 markers =
     slow: marks tests as slow (deselect with '--skipslow')
+
+# Execute tests in parallel.
+addopts = -n auto


.github/workflows/ci.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
           # OMP_NUM_THREADS: PySCF's poor OpenMP performance slows down QCQMC tests.
           # RECIRQ_IMPORT_FAILSAFE: skip tests on unsupported Cirq configurations.
           export OMP_NUM_THREADS=1
-          RECIRQ_IMPORT_FAILSAFE=y pytest -n logical -v --skipslow
+          RECIRQ_IMPORT_FAILSAFE=y pytest -v --skipslow
```
pavoljuhas, line 96: This may halve the number of parallel workers as `-n auto` selects the physical and `-n logical` the logical CPU count.  That said more workers may result in slower test run due to their startup and communication overhead so the best would be to just compare the CI execution times.

At any rate please include the `-n something` choice here so that the CI is not affected if pytest.ini is changed in the future.

pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
 markers =
     slow: marks tests as slow (deselect with '--skipslow')
+
+# Execute tests in parallel.
+addopts = -n auto
```
pavoljuhas, line 6: I am not sure about this.  Parallel run disables debug printouts from the code with `pytest -s` and also `pytest --trace some_test` runs.  That can be avoided by passing `-n0`, but that may be non-obvious for interactive use from shell.

(I'll approve if the change makes it easier on your side)

pavoljuhas (changes requested):
Please use explicit `-n auto` or `-n logical` (whichever is better on GHA) in ci.yaml.

pytest.ini
mhucka: Hmmm. The change in debugging is a good point, and probably not something that will be easy for other people to notice and deal with.

On balance, I think this PR is better not merged. I'll close it.

PR #470, opened 2026-04-15
Fix TKet error related to imports

dstrain115:
- TKet objects, such as Architecture, were being unresolved due to the fallback structure of the try/except blocks for ImportError.
- This PR cleans up the fallback structure and makes the dependency resolution for pytket a bit more robust.
- This should allow this file to be parsed by our documentation system.


mhucka (approved):
LGTM

PR #475, opened 2026-04-28
Fix Publication result download for Fermi-Hubbard

dstrain115:
- Data Dryad now requires an OAuth2 token.  This PR adds a parameter for the auth token and generates the publication notebook one-time using my auth token.  (Erased after generated the notebook).


recirq/fermi_hubbard/publication.py
```diff
@@ -235,29 +236,53 @@ def fetch_publication_data(
         base_dir: Base directory as a relative path to save data in.
             Set to "fermi_hubbard_data" if not specified.
         exclude: List of data to skip while downloading. See above for options.
+        auth_token: OAuth2 token for Dryad API.
     """
-    if base_dir is None:
-        base_dir = "fermi_hubbard_data"
+    if not os.path.exists(base_dir):
+        os.makedirs(base_dir)
```
pavoljuhas, line 242: Let us change the `base_dir` default above from `None` to `"fermi_hubbard_data"`.

recirq/fermi_hubbard/publication.py
```diff
@@ -235,29 +236,53 @@ def fetch_publication_data(
...
+            if response.status_code != 200:
+                print(f"Server returned {response.status_code}: {response.text}")
+                continue
+
+            with ZipFile(BytesIO(response.content)) as zfile:
                 zfile.extractall(base_dir)
+                print(f"Successfully downloaded and extracted {file_name}.\n")
+
+        except Exception as e:
+            print(f"Failed to download {file_name}: {e}")
```
pavoljuhas: This suppresses download and extraction errors - is that intended?

A notebook run will presumably fail later on due to missing files.

recirq/fermi_hubbard/publication.py
```diff
@@ -216,8 +216,9 @@ def rainbow23_layouts(sites_count: int = 8) -> Tuple[ZigZagLayout]:
 
 
 def fetch_publication_data(
-        base_dir: Optional[str] = None,
+        base_dir: Optional[str] = "fermi_hubbard_data",
```
pavoljuhas:
```suggestion
        base_dir: str = "fermi_hubbard_data",
```

pavoljuhas (approved):
Please see latest inline comments before merge.

PR #480, opened 2026-07-10
Fix massive overcommitment of resources when running pytest in parallel

mhucka:
Pytest on parallel systems was taking far more time to run than it should have. The problem turned out to be due to the fact that some of the ReCirq code used multiprocessing without taking account of the available CPU resources on the system. When running parallel pytests (with pytest-xdist), this resulted in the host system being vastly overcommitted.

This PR adds code to determine the number of available CPUs dynamically at run-time and control how many processes and threads are used. This is done in two ways:

* Change some classes that use multiprocessing directly (e.g., `src/openfermion/linalg/linear_qubit_operator.py`) to not use all processors always.

* Use the `threadpoolctl` package to limit how many threads native libraries such as BLAS use, so that packages like SciPy and NumPy don't end up using threads without control and without taking into account what pytest-xdist is doing.


recirq/benchmarks/xeb/grid_parallel_two_qubit_xeb.py
```diff
@@ -506,7 +506,8 @@ def compute_grid_parallel_two_qubit_xeb_results(
                 cycles,
             )
         )
-    num_processors = min(num_processors, len(arguments))
+    from recirq.testing_utils import get_available_cpu_count
```
dstrain115: Should this be at the top rather than inline?
Same below.

recirq/qcqmc/analysis_test.py
```diff
@@ -90,7 +90,7 @@ def test_small_experiment_partition_match_reversed_partition_same_seed(
         n_samples_per_clifford=n_samples_per_clifford,
         noise_model_name="None",
         noise_model_params=(0,),
-        seed=2,
+        seed=1,
```
dstrain115: What's up with this?

dstrain115 (approved):

recirq/benchmarks/xeb/grid_parallel_two_qubit_xeb.py
mhucka: Woops, yes, they should. Fixed in the next push.

Also, this made me realize the module name "testing_utils" is not that great for what it is, so I've renamed it to "cpu_utils". I hope that's okay?

recirq/qcqmc/analysis_test.py
mhucka: This is a fix to a problem revealed by another change in this PR. The other change is to pass the seed value to QSimSimulator [here in experiment.py](https://github.com/quantumlib/ReCirq/pull/480/changes#diff-ec12b7fe0fbf15b2a824f5b17e61093f15d257dbec08c75ad5a252f1ae6f3862R146). 

In this file (analysis_test.py), the `seed` parameter value for `experiment_2` _should_ be the same as the seed value for `experiment_1` (it's even implied by the test function's name, "test_small_experiment_partition_match_reversed_partition_same_seed"), but the problem was never caught because the `seed` parameter values in this test function have no effect unless the change to experiment.py is also made. The problem would never have been caught in CI either, because this particular test is marked "slow", and CI runs the tests with `--skipslow`.

Now thinking about it some more, the change to experiment.py is not actually necessary for the parallelism fix in this PR. I'll move that and this seed=1 change to another PR.

mhucka: Update: those two changes are now in PR https://github.com/quantumlib/ReCirq/pull/481

